PR #23, opened 2016-02-16
Merge fork-arnopaehler (e20834e)

Plaristote:
add QMLFont constructor in Checkbox and TextArea
change permitted versions in TextArea
add QMLAnimation constructor in SequentialAnimation

### DIFF
--- a/src/qtcore/qml/elements/QtQuick.Controls/Checkbox.js
+++ b/src/qtcore/qml/elements/QtQuick.Controls/Checkbox.js
@@ -7,7 +7,8 @@ registerQmlType({
     QMLItem.call(this, meta);
     var self = this;
 
-    this.font = new getConstructor('QtQuick', '2.0', 'Font')(this);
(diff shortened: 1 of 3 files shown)


ChALkeR:
Actual changes LGTM, but I think that there is no need to update `lib/` between releases, I opened #27 to discuss that.

Plaristote:
The question raised with me as well when I was working on these commits.
The lib folder was updated by arnopaehler, so I decided to follow the original committer's approach here. But I do agree that we shouldn't update lib/ on every commit (or even that lib/ shouldn't be in the repository at all).

From the current opinions in #27, for now we all agree that /lib shouldn't be added to every commit. Should I make another pull request in which these commits have been stripped of any changes in /lib ?

Plaristote:
I altered the pull request, removing all changes to the /lib folder.

ChALkeR:
Thanks!
LGTM, merging.

PR #24, opened 2016-02-17
Parse import statements using kde-master's qmlimport

Plaristote:
I patched the code from master so that it would load modules and javascript imports using the qmlimport parser form kde-master, instead of the one currently in master (which was written by myself a long time ago, and shouldn't be kept, as it is less concise and doesn't conform to the parsing logic used everywhere else).

I also removed the import of QML files (such as: import my.qml), since I couldn't get it to work right away, and didn't find anything in the QML documentation about these kind of imports.

The only missing thing from import is the recursive import of a directory containing qml files. Could be done for node.js users, though for web browsers, we can only do it using the qrc thingy (since we can't access the filesystem). We may want to open a low priority issue to add that feature in the future.

### DIFF
--- a/src/qtcore/qml/QMLEngine.js
+++ b/src/qtcore/qml/QMLEngine.js
@@ -107,7 +107,7 @@ QMLEngine = function (element, options) {
         // Create and initialize objects
         var component = new QMLComponent({ object: tree, parent: null });
         doc = component.createObject(null);
-        component.finalizeImports();
+        component.finalizeImports(this.rootContext());
         this.$initializePropertyBindings();
 
         this.start();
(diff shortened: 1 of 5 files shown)


ChALkeR:
> recursive import of a directory containing qml files

That should work through `qmldir`, and the code to do that should be in the kde branch. I do not know its current state, though, but will check that.

Plaristote:
Is this good for merging ?
It'll be more convenient to check the state of qmldirs with this implementation of qmlimport merged in.

ChALkeR:
This looks good to me, but I would prefer @akreuzkamp to re-check this, too.

aspidites:
@Plaristote,

Don't let this stop this PR from being merged, but I thought you'd be interested:
- [Directory Imports](http://doc.qt.io/qt-5/qtqml-syntax-directoryimports.html)

I assume that doing so addes a qmldir, and anything within a qmldir is importable.

akreuzkamp:
lgtm, approved :)

Issue #20, opened 2016-02-14
Merge fork-kde-realinheritance


ChALkeR:
Merge kde/realInheritance by @akreuzkamp.

- [ ] f676cc1 Make rootObject public via engine.rootObject.
- [ ] ebb1da4 Fix parsing of files with windows line-feeds.
- [ ] 043d693 Parse import statements correctly.
- [ ] e25251f Fetch xml data out of the dom sorted.
- [ ] 38b909a Remove warning when a signal is created on demand.
- [ ] 9b61c5b Use p instead of span inside Text elements.
- [ ] b3da103 Initialize states, transitions and transform properties.
- [ ] 11fb202 Don't use implicitWidth when width is binded, but not yet set.
- [ ] e30ac75 Fix property declarations with bindings.
- [ ] d7ae5de Fix cleanup when deleting objects and bindings.
- [ ] 5c90a47 Make XmlListModel.source loading asynchronous.
- [ ] 1d74538 Revert change Component.onCompleted to $finishInit for implicitSize.
- [ ] a99da86 Fix some context handling bugs.
- [ ] cd55b71 Do not try to add css-classes to non-Items.
- [ ] c98583d Use a list append method for ListModel instead of onItemsChanged.
- [ ] 22d307a Initial support for XmlListModel.
- [ ] be7bb90 Add support for keyboard handling.
- [ ] e8648d6 Add support for attached properties.
- [ ] f9dd070 Fix states. Makes a bad chromium bug appear.
- [ ] 4eefc48 Return binding-handling to the property object.
- [ ] f8acff6 Fix html widgets and move font css stuff from QMLText to QMLFont.
- [ ] de80982 Fix animation with an explicitly set from value of 0.
- [ ] a19d7e5 Fix implicit size of Images.
- [ ] a3b6d2b Fix Behavior.
- [ ] 7e5643e Fill list properties using an append function.
- [ ] b2424df Return to using an object per property.
- [ ] 17fd97f More progress.
- [ ] a5b4f4d Fix parsing of default properties.
- [ ] dc4ac1d Fix positioners.
- [ ] c481336 Fix scope handling with components.
- [ ] b9afeda Remove inheritance helper function.
- [ ] 8bd7e55 Fix implicit text size.
- [ ] e8018c8 Some progress.
- [ ] ad4645a Port some more types to the new syntax.
- [ ] 2443508 First draft.
- [ ] ec4bee9 Fix three small bugs with QMLListModel.
[...]

PR #28, opened 2016-02-17
Merge some of the commits from kde-master

akreuzkamp:
This review request includes 16 of 27(resp. 21 applicable) commits that were done to kde-master since @Plaristote did his fork.

### DIFF
--- a/src/qtcore/qml/elements/Component.js
+++ b/src/qtcore/qml/elements/Component.js
@@ -76,6 +76,7 @@ function QMLComponent(meta) {
         this.finalizeImports(this.$context);
     }
 }
+inherit(QMLComponent, QMLBaseObject);
 
 registerQmlType('Component',   QMLComponent);
 registerQmlType('QMLDocument', QMLComponent);
--- a/src/qmlweb/RestModel.js
+++ b/src/qmlweb/RestModel.js
@@ -2,6 +2,7 @@ registerQmlType({
   module:   'QmlWeb',
   name:     'RestModel',
   versions: /.*/,
+  baseClass: QMLItem,
   constructor: function QMLRestModel(meta) {
     QMLItem.call(this, meta);
     var self = this;
--- a/src/qmlweb/Settings.js
+++ b/src/qmlweb/Settings.js
@@ -2,6 +2,7 @@ registerQmlType({
   module: 'QtQuick',
   name:   'Settings',
   versions: /.*/,
+  baseClass: QMLItem,
   constructor: function QMLSettings(meta) {
     QMLItem.call(this, meta);
 
--- a/src/qtcore/qml/QMLPositioner.js
+++ b/src/qtcore/qml/QMLPositioner.js
@@ -8,6 +8,7 @@ function QMLPositioner(meta) {
 
     this.spacing = 0;
 }
+inherit(QMLPositioner, QMLItem);
 
 QMLPositioner.slotChildrenChanged = function() {
     for (var i = 0; i < this.children.length; i++) {
--- a/src/qtcore/qml/elements/Item.js
+++ b/src/qtcore/qml/elements/Item.js
@@ -389,5 +389,6 @@ function QMLItem(meta) {
         }
     }
 }
+inherit(QMLItem, QMLBaseObject);
 
 constructors['Item'] = QMLItem;
--- a/src/qtcore/qml/elements/PropertyChanges.js
+++ b/src/qtcore/qml/elements/PropertyChanges.js
@@ -16,5 +16,6 @@ function QMLPropertyChanges(meta) {
         });
     }
 }
+inherit(QMLPropertyChanges, QMLBaseObject);
 
 registerQmlType('PropertyChanges', QMLPropertyChanges);
--- a/src/qtcore/qml/elements/QtGraphicalEffects/FastBlur.js
+++ b/src/qtcore/qml/elements/QtGraphicalEffects/FastBlur.js
@@ -2,6 +2,7 @@ registerQmlType({
   module:   'QtGraphicalEffects',
   name:     'FastBlur',
   versions: /.*/,
+  baseClass: QMLItem,
   constructor: function QMLFastBlur(meta) {
     QMLItem.call(this, meta);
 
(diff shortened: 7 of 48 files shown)


Plaristote:
I see we worked on the same thing this morning. Luckily we only merged different commits, except for a3e8bf1.
I'll stop working on this branch for now, you can check #29 to see which commits I've already worked on.

ChALkeR:
This introduces an LGPL-licensed file into the bundle,  without mentioning it in the actual LICENSE. This should be fixed in some way.

ChALkeR:
Another note — can we have meaningful commit messages, please? =)

ChALkeR:
@akreuzkamp Could you rebase this, please, and make the commits a bit more clear (i.e. without the prefix).

ChALkeR:
02822c9 — Merged in #29.
3da78c7 — Merged in #29.
785965c — Merged in #29.
3fab595 —
1c12ea4 — Merged in #29.
d35f213 — Merged in #29.
c3f1f39 — Merged in #29.
12b725e — Merged in #29.
4c9e432 — Merged in #29.
48db468 — Merged in #29.
6022424 —
79a05ff —
2a7202f —
7dad8e8 —
f401c53 —
44bea45 —

ChALkeR:
I'm rebasing this and excluding the already merged commits.

ChALkeR:
The tests fail with:
```
TypeError: component.createObject is not a function
    at loadQMLTree (QMLEngine.js:120)
    at loadQML (QMLEngine.js:103)
    at loadQml (common.js:17)
```

ChALkeR:
Looks to happen mainly because of `QMLBaseObject`.

src/qtcore/qml/elements/Component.js
```diff
@@ -76,6 +76,7 @@ function QMLComponent(meta) {
         this.finalizeImports(this.$context);
     }
 }
+inherit(QMLComponent, QMLBaseObject);
```
ChALkeR: Most of the fails happen because of this change.

ChALkeR:
The problem is that `inherit` overrides the prototype _after_ some methods have been set on the prototype.

ChALkeR:
@akreuzkamp, please review my last commit here, it fixes things.

ChALkeR:
The changes here are pretty straightforward.

First commit moves TextArea (from QtQuick.Controls) to QtQuick as TextEdit, and makes TextArea from QtQuick.Controls an alias for that (both are multi-line text input fields). This seems reasonable to me, it gives us a TextEdit implementation for now, which is important. This should be improved in the future, and differences between TextEdit and TextArea should be implemented, but this is out of scope for now.

Second commit implements prototype-based inheritance, which is also a good thing.

Third commit (the one by me) just fixes the second one so that manually assigned prototype methods that were called _before_ `registerQmlType` do not get lost.

There are still more work that needs to be done (but I would prefer to see that in separate PRs), e.g. `Button` is not described correctly, but it's not the fault of this commit. See #48 for details. Another one would be to re-check manual calls of constructor methods and (most likely) removal of most of those.

@Plaristote, @henrikrudstrom, could someone please give this a quick review? I want to land this on master.

Btw, it's the port of  6022424 that you were discussing here: https://github.com/qmlweb/qmlweb/issues/20#issuecomment-193214035.

ChALkeR:
Hm. On the other hand, this doesn't include `getAttachedObject` from 6022424 for whatever reason, so the commit message should be changed. This is just the inheritance.

Plaristote:
Everything looks good to me here, though I'm wondering about something in these kind of changes:
```javascript
registerQmlType({
  module:   'QtQuick',
  name:     'State',
  versions: /.*/,
  baseClass: QMLBaseObject,
  constructor: function QMLState(meta) {
    QMLBaseObject.call(this, meta);
```
The redundancy of QMLBaseObject is unsettling to me. I can live with it, I'm sure that trying to change this would go beyond the scope of this PR. But it sounds to me like everything done by `QMLBaseObject.call(this, meta)` should be done by checking the baseClass sent to registerQmlType.

ChALkeR:
@Plaristote Yes, I was talking about those in https://github.com/qmlweb/qmlweb/pull/28#issuecomment-193765917:
> Another one would be to re-check manual calls of constructor methods and (most likely) removal of most of those.

I will file a follow-up PR to clean those things up a bit.

ChALkeR:
@Plaristote I opened #133 so we don't forget about that =).

ChALkeR:
Argh. Forgot to amend the commit message. Force-push time!

akreuzkamp:
Ok, I think I missed quite some discussion here...
1. Missing getAttachedProperty() method: That got lost during the merge somehow. We need to readd that to the commit.
2. redundant constructor calls: I don't see a way to reduce that. `baseClass: QMLBaseObject` alters the prototype and thus makes some inherited methods available. So it alters the "class" QMLBaseObject. `QMLBaseObject.call(this, meta);` calls the base constructor. It alters the object and needs to be done everytime we create an object, not only once when we register the type. Those two lines do fundamentally different things. I don't see a way, removing one of them. Or does javascript provide a way to call base constructors automatically and I just don't know that?
3. @ChALkeR's fix: I'm not sure... What you did, seems like a workaround to me. Shouldn't we instead just fix the broken things? To me the reason for the failures is simple: inherit was called to late. It needs to be called before adding things to the prototype.

Plaristote:
Wouldn't something like that be ok ?
```js
function registerQmlType(options) {
 // ...
  var constructor = function(meta) {
    options.baseClass.call(this, meta);
    options.constructor.call(this, meta);
  }
}
```

ChALkeR:
@akreuzkamp 
> @ChALkeR's fix: I'm not sure... What you did, seems like a workaround to me. Shouldn't we instead just fix the broken things? To me the reason for the failures is simple: inherit was called to late. It needs to be called before adding things to the prototype.

I don't think that elements defining their methods on the prototype should be considered "broken", they were just not compatible with this new change. Atm, this change allows more flexibility with how the modules are defined without any significant drawbacks. We could reconsider that and refactor later, I don't think that we should refactor things before everything else is merged.

akreuzkamp:
@Plaristote: Yes, ok, that works of course :)

@ChALkeR: I don't think that, either. ;)
The only thing, that is required, is to call inherit first. :) Your solution is a bit too "automagic" for my taste, because it moves a lot of stuff around. I'd prefer to just require inherit to be called first.
But you're right about the "merge things first." We can refactor that later. :)

PR #29, opened 2016-02-17
Merge kde master

Plaristote:
Merges a good deal of commits from kde-master to master.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,11 @@
 node_modules/
-tmp/
 lib/
+tmp/
+build/
 *~
 *.bak
 .*.swp
 .*.swo
+*.kdev4
+*kate-swp
+.idea
--- a/src/qtcore/qml/QMLBaseObject.js
+++ b/src/qtcore/qml/QMLBaseObject.js
@@ -5,7 +5,7 @@ function QMLBaseObject(meta) {
         prop;
 
     if (!this.$draw)
-        this.$draw = noop;
+        this.$draw = function(){};
 
     if (!this.$isComponentRoot)
         this.$isComponentRoot = meta.isComponentRoot;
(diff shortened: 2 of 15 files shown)


.gitignore
```diff
@@ -1,6 +1,10 @@
 node_modules/
 tmp/
+build/
```
ChALkeR: Not sure where the `build/` comes from.
@akreuzkamp, why do we want this in `.gitignore`?

henrikrudstrom:
Looks good to me!

Plaristote:
I'd merge this with your reviewing, but I'd like to hear @akreuzkamp 's answer about the build/ directory. I'm wondering if it has something to do with code that we dropped.

ChALkeR:
Other things (except the `build/` in gitignore, but that is minor) seem good, but, again, I would prefer if @akreuzkamp took a quick look on these commits.

.gitignore
akreuzkamp: qmlweb2 creates it ;)
I think .gitignore is a place to ask "why not?", rather than "why?", because what ever helps one developer should go in ;)
But feel free to remove, if you don't agree :)

akreuzkamp:
approved! :)

akreuzkamp:
let's merge this, it doesn't break my code and seems to be slightly better integrated into your code, than my merges. :)

.gitignore
ChALkeR: Ok, seems reasonable =).

ChALkeR:
Merging.

ChALkeR:
There were some conflicts in `Add method Qt.createComponent similar to Qt's one.` and `Make rootObject public via engine.rootObject.`, I ameneded them.

I also renamed `Added .gitignore` to `Updated .gitignore`.

ChALkeR:
Landed.

PR #31, opened 2016-02-17
jsparser requires uglify-js if nodejs is detected

Plaristote:
Introduces a condition, making our jsparse function use a globally defined "parse" function when run from a web browser, and requiring uglify-js when run by nodejs.

This change is required for the `gulp-qmlweb` package to work.


ChALkeR:
LGTM, merging.

Note that this is not the only required change, though — to be useful as an module we should actually export something.

PR #36, opened 2016-02-18
package.json improvements: explicitly list files, add repo field.

ChALkeR:
Explicitly listing files is one of two ways of overriding `.gitignore` if we put `lib/` there, another one would be to maintain an `.npmignore` that is equal to `.gitignore` minus the `lib/`, but that would be inconvenvient and would affect the `.gitignore` merges, perhaps.

This also ensures that no unexpected files are packaged (note that there are no directories listed, only glob patters with file extensions).

`tests` and `src` are packages purposely, I think that those should be always coupled with the built lib, also note that `src` could be used directly atm.

Including the `repository` field should give us a link to the repository, issues, and pull requests at the npm package page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.4",
   "description": "QML port to Javascript",
   "license": "BSD",
+  "repository": "qmlweb/qmlweb",
   "devDependencies": {
     "gulp": "^3.6.2",
     "gulp-concat": "^2.1.7",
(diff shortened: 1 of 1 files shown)


Plaristote:
I wonder if we really need to import qml files from `src/`, but I assume that in the future, we may provide some elements written directly in qml, and they'd be stored in src. If that's the point, everything here looks good to me.

ChALkeR:
@Plaristote Yes, that was the plan — to split the compiler and elements, and to write some elements in QML.

Merging.

Issue #27, opened 2016-02-17
`lib/` status


ChALkeR:
We need to decide what to do with the `lib/` dir (which contains bundled code).

Do we keep it (and update it) or do we remove it from the repo?
If we keep it — how often do we update it? With each commit or between the releases?

Options that I see (anyone could propose any other solution):
 1. Remove `lib/` from the repo. Downsides — it would be harder to download this directly from GitHub, and I guess this will make the package unusable with Bower. I could be mistaken here, though.
 2. Update `lib/` before each release tag, with a separate commit. Downsides — `lib/` in the repo is desynced with the actual changes in the code, and this would make everyones `git status` not so pretty — there will (almost) always be uncommited changes in the `lib/` dir.
 3. Update `lib/` with each commit. Downsides — this would make merging, reverting, and every other everyday work a little more inconvenient. The work-around is to reset and rebuild it every time, but this would still break automatic merges.
 4. Remove `lib/` from the master branch (as in 1), make a separate branch (or branches if we would want to keep several of those somewhere in the future) for releases and package all releases from that branch (as in 2). Downsides — `master` would be uninstallable with Bower, but this looks better than having `master` actually install a previous version.
5. Remove `lib/` from the master branch (as in 1), make tags for each release that are not inside any branches, but in individual tags. Looks similar to 4, see a follow-up comment.

/cc @qmlweb/core

ChALkeR:
If there are no other ideas, I vote for option 2.

aspidites:
Another idea from a random guy:
4. Get rid of it, but include it in released tarballs. The end result would be that individuals who just want to try out the project would do so with code that is in sync with everything else that was released, and those that want bleeding edge would simply have to build the files for themselves. It also avoids the possibility of the code that is forked being out of sync with what is provided in /lib.

henrikrudstrom:
I like random guys suggestion. Less room for confusion!

Plaristote:
I would be in support for option 1. At least for now, because we're not ready to make ourselves available to bower users anyway.

I believe option 4 from aspidites doesn't fix the bower issue, but we could, instead of a tarball, just make a different repository to publish the releases.

aspidites:
You could have master always refer to latest release, do development on feature branches, and have a tagged branch for each release that has the associated files (hybrid of what is now listed as 4 in the OP, my original suggestion, and @Plaristote's comments). 

In effect, you'd have the following scenarios:

[...]

Issue #4, opened 2016-02-14
Add authors to the org


ChALkeR:
All collaborators should be probably asked to join to the org.
Will prepare a list a bit later, should be based on #3.

ChALkeR:
The reasoning here is that the current situation with a lot of forks at various states and almost all of them being quite old and not actively worked on saddens me. Hence I wanted to bring everyone into one org, which would be governed by the collaborators.

henrikrudstrom:
Great initiative, not much to see in my fork, but i hope to be able to contribute here!

ChALkeR:
Everyone, there are two teams — Core (@qmlweb/core) and Collaborators (@qmlweb/collaborators).

The aim is to have functional code review (each commit is reviewed by at least one other Collaborator), but that could happen not so soon as I hope it would.

The decision making would be consensus-seeking, first among all Collaborators, and if that fails — among the Core team.

These people are invited to the Core team:
 * @lpaimen
 * @akreuzkamp 
 * @Plaristote 
 * @arnopaehler

These people are invited to the Collaborators team:
 * @labsin 
 * @Somsubhra 
 * @henrikrudstrom 
 * @JoshuaKolden 

Ping me here to get the actual org invitation

lpaimen:
Good initiative! I've seen a lot of forks which tells me people find Qmlweb useful or at least want to tinker with it. Having a single "effort" makes it much smoother. I'm happy you picked this up!

labsin:
Yes, I was also disappointed that there were only dead forks at the time. But I also had some fundamental issues with the program. I tried to start from scratch (not released) but also gave up.

ChALkeR:
@labsin I agree that there are issues, but I think that it would be possible to do a gradual rewrite without starting from scratch.

Plaristote:
Thanks for the initiative of organizing us and all the work put towards integrating our respective repositories. I'll gladly accept the invitation. I'm always eager to work on this project, I see a lot of potential in it.

I'll check what I can do to help merging everyone's work here. It's very exciting to see everyone's work, even more-so when I see all that's already been merged.

akreuzkamp:
Hey Chalker,
thanks a lot for picking this up again! :)

I tried to pull us all together once, but failed, so I'm very glad to see you having more success.

A few weeks ago, I started to merge all my work into Plaristote's fork (locally still), because I think it's superior to mine. Please add me as a member to the org, so I can contribute the merge. :)

[...]

PR #44, opened 2016-02-19
Update README

aspidites:
I wanted it to be a bit clearer that the /lib directory was generated from gulp, so added a touch of formatting to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,12 +31,32 @@ This project aims at bringing the power of QML to the web browser.
 
 ## How to use
 #### Add the library to your web page
+Using one of the methods below, install the qmlweb JavaScript library:
 
-Install with [npm](https://www.npmjs.com/package/qmlweb),
-[Bower](http://bower.io/search/?q=qmlweb), or download directly from the
-GitHub [releases](https://github.com/qmlweb/qmlweb/releases) page.
+- [npm](https://www.npmjs.com/package/qmlweb):
 
-From that package, preload the file `lib/qt.js` in an HTML page.
+  ```
+  npm install qmlweb
+  ```
+
+- [Bower](http://bower.io/search/?q=qmlweb):
+
+  ```
+  bower install qmlweb
+  ```
+
+- GitHub [releases](https://github.com/qmlweb/qmlweb/releases):
+  ```
+  tar -xzvf v0.0.4.tar.gz
+  ```
+
+- Manually using gulp (recommended if you cloned from git):
+  ```
+  npm install
+  npm run build
+  ```
+
+Next, simply add `lib/qt.js` to the list of other JavaScript files in your app's HTML file:
 
 
 ```HTML
@@ -47,7 +67,15 @@ From that package, preload the file `lib/qt.js` in an HTML page.
 You may then modify the `<body>` element to specify what QML file to load when the page is opened.
 
 ```HTML
-<body style="margin: 0;" data-qml="qml/main.qml">
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>QML Auto-load Example</title>
+  </head>
+  <body style="margin: 0;" data-qml="qml/main.qml">
+    <script type="text/javascript" src="/lib/qt.js"></script>
+  </body>
+</html>
 ````
 
 ## How to use with Gulp


henrikrudstrom:
More user friendly, always good!

ChALkeR:
@aspidites This should be rebased against `master` (after #27) got landed, and I suppose that I have some further comments about this.

aspidites:
@ChALkeR updated with upstream master. Comment away!

Plaristote:
@aspidites I'm not sure that's what Chalker expected. Rebasing is different from merging. Instead of `git merge master`, you would do `git rebase master`, fix each conflicts that occurs, then go `git rebase --continue` until you're done.
I believe you can still do this, using the reset command, then rebase, then push with the force option.

I have a few comments as well: firstly, I'm fairly certain it's written "Javascript", not "JavaScript"; secondly, I sure I'm the author of this one, but the wording seems wrong on this sentence:
```
From that package, preload the file `lib/qt.js` in an HTML page.
```

aspidites:
> I have a few comments as well: firstly, I'm fairly certain it's written "Javascript", not "JavaScript"; 
Firefox's English dictionary suggests "JavaScript" when I try to type "Javascript". I also get this when googling for "javascript":

2016-02-23-193239_641x318_scrot


> @aspidites I'm not sure that's what Chalker expected. Rebasing is different from merging.

I know how to rebase. Didn't realize he wanted a literal rebase. Can do though.

>  I sure I'm the author of this one, but the wording seems wrong on this sentence

I definitely didn't write it :-P. What about something like this:

> Next, simply add `lib/qt.js` to the list of other JavaScript files in your app's HTML file

aspidites:
@Plaristote, I went ahead and did a rebase. Waiting on feedback about the other two points (proper capitalization and wording) before I make further changes.

henrikrudstrom:
lgtm

ChALkeR:
Not sure why the script has to be put in the body tag nowdays. Could you elaborate?

aspidites:
@chALkeR

It doesn't have to be. It's preferred nowadays because putting it in head then means you need to include your JavaScript is run after the page is fully loaded, which means listening to the document's onready event. As the body is the last element to be loaded, putting JavaScript at the end of it allows you to not worry about this. I think it might also be caster in some cases.

I can find some more official references if you like.

Plaristote:
I can't find anything about whether the onload/onready events can trigger before all the javascript files are loaded when they're put at the bottom of the body.
If that's true, we should probably work on the AutoLoader before suggesting of putting the javascript in the body tags. Otherwise, we should indeed suggest to put the javascript at the bottom of the page (see: https://developer.yahoo.com/performance/rules.html#js_bottom).

aspidites:
Also note this is also recommended by [w3schoolks](http://www.w3schools.com/js/js_whereto.asp):
> It is a good idea to place scripts at the bottom of the <body> element.
This can improve page load, because script compilation can slow down the display.

aspidites:
@Plaristote any comments on what I mentioned above about the capitalization of js and the rewording of that section?

Plaristote:
I like the sentence you reworded, sounds more appropriate to me, I'm all for it.
And you were right about the capitalization as well.

aspidites:
@Plaristote updated with your suggestions. Sorry if I've sounded a bit curt in my earlier responses. Often I've been either typing from a phone or in between "real" work (AKA, that thing that lets me keep my house :-P).

I appreciate everyone's feedback so far.

ChALkeR:
> Also note this is also recommended by w3schools:

Never listen to w3schools. Btw, they are not associated with W3C in any way. Also note that they have a lot of mistakes and/or outdated info.

Ref: http://www.w3fools.com/

henrikrudstrom:
Regarding script tag at the end: since we generate the HTML with javascript, there wont be much html to show before qt.js is loaded...

ChALkeR:
> It doesn't have to be. It's preferred nowadays because putting it in head then means you need to include your JavaScript is run after the page is fully loaded, which means listening to the document's onready event. As the body is the last element to be loaded, putting JavaScript at the end of it allows you to not worry about this. I think it might also be caster in some cases.

https://developer.mozilla.org/ru/docs/Web/HTML/Element/script#Browser_compatibility
There is an `async` attribute for that now.

aspidites:
`async` only works on IE >= 10. Seems like a silly reason to need a shim.

Is the qml data attribute always going to be on the body tag? Is it not conceivable that someone will want to embed a qml app inside some bootstrap (for instance) scaffolding?

ChALkeR:
Also I'm not sure what value do the expanded html element bring in the first case — that html doesn't do anything. Perhaps just leave `<script type="text/javascript" src="/lib/qt.js"></script>` as is there?

As for the second case (auto-loading) — providing a full HTML example looks like a good idea there, but the HTML should be at least valid. Atm, it doesn't validate here: https://validator.w3.org/nu/#textarea

aspidites:
I gave full HTML examples so that at any point in the readme you could just copy/paste and go. I'll update the snippets when I'm not at work.

Plaristote:
> Regarding script tag at the end: since we generate the HTML with javascript, there wont be much html to show before qt.js is loaded...
True, though some people may want to display a loader, while the javascripts are being fetch.
The `async` attribute works in this case as well.

I personally like to keep things in the head, and use `async` when I can afford it. However, nowadays, it's much more common to just put the javascript at the bottom of the page.

aspidites:
I removed the superfluous HTML from the first example, and made the second valid HTML. I originally included meta charset, but decided to go with the minimal validating example that @ChALkeR suggested. 

I also kept the script tag at the end as it's common practice and the vote seems to be essentially 50/50 with me being +1, @ChALkeR being -1, and @Plaristote being +0 (preferring async but recognizing tradition). Happy to be told otherwise and make that change as well if necessary.

ChALkeR:
@aspidites 
> and the vote seems to be essentially net 0 

For the reference, I don't think that we should be _adding up_ votes at this point, see https://github.com/qmlweb/qmlweb/issues/4#issuecomment-184008678:
> The decision making would be consensus-seeking, first among all Collaborators, and if that fails — among the Core team.

Consensus-seeking is described here, for example: https://en.wikipedia.org/wiki/Consensus-seeking_decision-making.

--

Speaking of this PR, I think that I'm fine with `<script>` tag being in the `<body>` in that example for now. That could be discussed in a separate issue, if needed.

Could you squash the commits, please?

ChALkeR:
Btw, should we replace `lib/qt.js` with `lib/qt.min.js`?

README.md
```diff
@@ -31,12 +31,32 @@ This project aims at bringing the power of QML to the web browser.
...
+
+- GitHub [releases](https://github.com/qmlweb/qmlweb/releases):
+  ```
+  tar -xzvf v0.0.4.tar.gz
+  ```
+
+- Manually using gulp (recommended if you cloned from git):
+  ```
+  npm install
+  gulp build
```
ChALkeR: Use `npm run build` instead of `gulp build`.

It launches `gulp`, but does not require `gulp` to be installed globally for building `qmlweb`.

aspidites: I originally opened this PR because it wasn't obvious where lib/qt.js came
from. It was not to bikeshed for a week about where a script tag should go.
I'd just as well reverse my addition of HTML considering how surprisingly
controversial that has been.

ChALkeR: @aspidites Sorry, I didn't want to bikeshed, just want everything to be as good as possible =).

Note that the comment is not about the HTML (we already agreed on that, I think), but about replacing `gulp build` with `npm run build`, because `gulp build` will fail for users that don't have `gulp` installed globally, and we don't want that, correct?

aspidites: Sorry about that. I replied via email and was only meaning that if the HTML changes were that controversial I could go back to what was originally there. 

I totally agree with the gulp changes.

henrikrudstrom:
LVGTM

ChALkeR:
LGTM

PR #47, opened 2016-02-19
TextInput: fallback to password for NoEcho, PasswordEchoOnEdit

ChALkeR:
This is done instead of merging 8c6ec044d51cdd8cc19bd7e521616149bc4f6f09, because most part of that is already implemented in a different way.

### DIFF
--- a/src/qtcore/qml/elements/QtQuick/TextInput.js
+++ b/src/qtcore/qml/elements/QtQuick/TextInput.js
@@ -49,6 +49,14 @@ registerQmlType({
           case TextInput.Password:
             this.dom.firstChild.type = "password";
             break ;
+          case TextInput.NoEcho:
+            // Not supported, use password, that's nearest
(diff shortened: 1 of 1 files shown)


Plaristote:
Looks good to me

henrikrudstrom:
LGTM
@Plaristote im open for `options` or adding a 5th parameter, but that should be a separate PR

akreuzkamp:
lgtm

Plaristote:
LGTM.
PR #78 can make the `options` a reality if we all agree that's how createSimpleProperty should behave.

ChALkeR:
@Plaristote I don't see anything in this PR that uses the default parameter.

ChALkeR:
@Plaristote Ah, actually that was in the initial commit by @akreuzkamp. I see. Missed that, thanks for noticing!

PR #50, opened 2016-02-19
Use karma, add Coveralls integration.

### LINKED ISSUE #40: Code coverage analysis
We should probably integrate with [Coveralls](https://coveralls.io/github/qmlweb/qmlweb) to get code coverage analysis reports and badge.

Istanbul should generate the coverage output files.

ChALkeR:
This gives us crude coverage analysis.
It's based on a single file atm (`lib/qt.js`), that has to be fixed later.
Another related problem is that it also analyzes the bundled `uglify-js` code.

There are several ways of fixing that, but that could be done in follow-up PRs.

This (sort of) fixes #40.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules/
 lib/
+coverage/
 tmp/
 build/
 *~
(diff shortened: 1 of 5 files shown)


ChALkeR:
/cc @qmlweb/collaborators I want this reviewed.
It doesn't give us detailed analysis yet (that will follow), but it does give us an overall percentage.
Some more info is available in the `coverage` folder after running `npm test` locally.

henrikrudstrom:
That worked much better, had trouble getting gulp-jasmine-phantom to work, and it didnt seem so active.

ChALkeR:
@henrikrudstrom Should I treat this as a review? =)

ChALkeR:
Note that internally this still uses the same testing framework, Jasmine.

henrikrudstrom:
if you want. Im no expert on testing, but this works out of the box and seems much more active, only thing i miss is the green text saying which tests passed...

ChALkeR:
No objections, merging.
This is just toolchain changes and do not break the actual code, so I'm confident in this.

ChALkeR:
@henrikrudstrom 
> only thing i miss is the green text saying which tests passed...

When there will be many tests, it would be noisy. It outputs tests that fail =).

PR #53, opened 2016-02-20
Add basic elements tests

ChALkeR:
This requires #50 and is based upon it, so that should land first.

This increases our tests coverage from 12.11% to ~~23.92%~~ 24.49%. =)

PR #54, opened 2016-02-20
tests: Add render tests support

ChALkeR:
This branch requires #53 and is based on it.

This commit adds basic render tests support.

No js required for new tests, all paired `.qml`/`.png` files from the `tests/Render/` directory subdirectories are treated as render tests. This renders each `.qml` file from there and compares it with a corresponding `.png` file.

Running with `QMLWEB_SAVE_RENDER=1` env var will also dump all the renders to the `./tmp/render` dir (created automatically, if needed).


ChALkeR:
No objections raised, merging.
This is just toolchain change and does not break (or touch) the actual code, so I'm confident in this.

PR #55, opened 2016-02-20
Language tests and context fix 

henrikrudstrom:
Added some simple language tests, found an error where only the root object had its $context property set, but not its children. 
applied 9d4c4cf14fded5c4dcbc499fb7f4d862a66a988e from kde-master to make the test pass.

also added gulp task karma-debug that launches a debugger in chrome/chromium

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules/
 lib/
+coverage/
 tmp/
 build/
 *~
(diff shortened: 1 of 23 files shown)


ChALkeR:
This should be rebased now.

henrikrudstrom:
replaced by #63

PR #56, opened 2016-02-20
Merge: Don't override inner component context.

henrikrudstrom:
> The construct function overrided the inner context with some empty
context. This was due to an earlier misunderstanding of how the
contexts work and was completely meaningless (except of causing bugs).
I simply removed that line now.


ChALkeR:
LGTM, merging.

ChALkeR:
Merged, thanks!

PR #57, opened 2016-02-20
Kde master ongoing merge — rebased

ChALkeR:
This is a rebased version of #28.

Merged commits: 79a05ff, f401c53, 7dad8e8, 44bea45.

### DIFF
--- a/src/qtcore/qml/QMLProperty.js
+++ b/src/qtcore/qml/QMLProperty.js
@@ -103,7 +103,11 @@ QMLProperty.prototype.set = function(newVal, fromAnimation, objectScope, compone
             }];
             this.animation.running = true;
         }
-        this.changed(this.val, oldVal, this.name);
+        if (this.obj.$syncPropertyToRemote instanceof Function && !fromAnimation) { // is a remote object from e.g. a QWebChannel
+            this.obj.$syncPropertyToRemote(this.name, newVal);
+        } else {
+            this.changed(this.val, oldVal, this.name);
+        }
     }
 }
 
--- a/src/qtcore/qml/lib/parser.js
+++ b/src/qtcore/qml/lib/parser.js
@@ -269,11 +269,12 @@ function parse_js_number(num) {
         }
 };
 
-function JS_Parse_Error(message, line, col, pos) {
+function JS_Parse_Error(message, line, col, pos, comment) {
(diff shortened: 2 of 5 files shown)


ChALkeR:
Actually, 9ebf5d8 is the first failing commit, something is wrong with it.

ChALkeR:
@akreuzkamp Thanks for the update!
adacdf8 looks fine now, but c9d178c is breaking the build.

Plaristote:
We may want to write some test to make this work. I've checked the bindingArray commit, and it doesn't work with me: an array that uses bound values just gets evaluated as `undefined` with me.

I've tested ce1c580, it LGTM.
I'm not sure how to check b5a33b8.
The array binding however doesn't seem to work.

Plaristote:
I was wrong about the array binding. My test environment must've had issues: I wrote a test to check it, and it does work.
Check PR #73 to see the test. It currently breaks, and that's normal: it works if it is rebased on `kde-master-ongoing-merge-rebased`... so it'll be fine once this is merged in master.

henrikrudstrom:
looks hard to test b5a33b838e99086eab54d8af8d168fa6a8f06df5 without an actual webchannel to test with.
added a test for ce1c580c7bbbf9e0a52e2dc8fd069523e513d936 but it doesnt seem to report the correct line number.

ChALkeR:
Squashed the test and rebased.

ChALkeR:
Fixed the test syntax.

ChALkeR:
@pavelvasev It looks like the `line` starts with `0` in your code, while @henrikrudstrom expects it to start with `1`. That's what fails the test. How should this behave?

henrikrudstrom:
@ChALkeR @Plaristote @akreuzkamp 
shall we?

ChALkeR:
Ok, I'm going to split this today and merge some of those commits.

ChALkeR:
Actually, it doesn't split without dependencies.

The new test looks good to me.
The parse error and the parse error fix also look good to me.
The parser fixes are covered by the new test in #73, so rubber-stamp LGTM.
«Add basic support for QtWebChannel.» just adds some hooks that seem required for some usecases, and that should not break anything. Rubber-stamp LGTM.

I'm going to merge this.

PR #59, opened 2016-02-21
Reduce unneeded evals

ChALkeR:
This makes the execution scope cleaner — `eval()` captures the parent scope, `new Function()` doesn't.
A testcase to ensure that `bindSrc` not being exported to the binding scope is included (fails on master).

/cc @Plaristote, @akreuzkamp


igosoft:
Can we expect performance change ?

ChALkeR:
@igosoft I don't think so, but you can test this =).

ChALkeR:
@igosoft The short-term plan is to merge everything, increase tests coverage, and fix compatibility issues.

akreuzkamp:
lgtm, +1

ChALkeR:
Landed.

Issue #21, opened 2016-02-14
eslint all the things


ChALkeR:
eslint should be added to gulpfile, but it would be better to do that after #15 gets done.

PR #64, opened 2016-02-21
Enable eslint for the ./tests folder

ChALkeR:


ChALkeR:
/cc @Plaristote @akreuzkamp — which format do you prefer? Any considerations? Proposals? Should we discuss some of these rules or should we add some more?

I plan to extend these to all the sources once the merge would be completed (#21).

henrikrudstrom:
@ChALkeR unless anyone has strong opinions here, which it looks like there arent, how about we enable it for the tests now, just to test the linting, we can alway tweak later

ChALkeR:
@henrikrudstrom Should I take that as an «LGTM»?

henrikrudstrom:
LGTM

ChALkeR:
Rebased.
I also fixed lint errors at other places and removed the `no-underscore-dangle` rule, it was broken by the new `Initialize` tests (those have `_version` and `_depends` properties).

ChALkeR:
Landing.

PR #65, opened 2016-02-21
test task tweaks

henrikrudstrom:
Made some changes to the test gulp tasks:
* dont rebuild qt.min.js on every test build
* Karma seems to magically rebuild the qt.js file so i renamed karma to watch-tests and removed the latter (please verify)
* added chrome/chromium debugging

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -35,7 +35,7 @@ var tests = [
   'tests/**/*.js'
 ];
 
-gulp.task('qt', function() {
+gulp.task('build-dev', function() {
   return gulp.src(qtcoreSources)
              .pipe(order(qtcoreSources, { base: __dirname }))
              .pipe(sourcemaps.init())
@@ -45,7 +45,7 @@ gulp.task('qt', function() {
              .pipe(gulp.dest('./lib'));
 });
 
-gulp.task('min-qt', ['qt'], function() {
+gulp.task('build', ['build-dev'], function() {
   return gulp.src('./lib/qt.js')
              .pipe(rename('qt.min.js'))
              .pipe(changed('./lib'))
@@ -55,7 +55,13 @@ gulp.task('min-qt', ['qt'], function() {
              .pipe(gulp.dest('./lib'));
 });
 
-gulp.task('build', ['qt', 'min-qt']);
+gulp.task('watch', ['build'], function() {
+  gulp.watch(qtcoreSources, ['build']);
+});
+
+gulp.task('watch-dev', ['build-dev'], function() {
+  gulp.watch(qtcoreSources, ['build-dev']);
+});
 
 gulp.task('test', ['build'], function(done) {
   new karma.Server({
@@ -68,16 +74,18 @@ gulp.task('test', ['build'], function(done) {
   }).start();
 });
 
-gulp.task('karma', ['build'], function(done) {
+gulp.task('test-watch', ['watch-dev'], function(done) {
   new karma.Server({
     configFile: __dirname + '/karma.conf.js'
   }, done).start();
 });
 
-gulp.task('watch', ['build'], function() {
-  gulp.watch(qtcoreSources, ['build']);
+gulp.task('test-debug', ['watch-dev'], function(done) {
+  new karma.Server({
+    configFile: __dirname + '/karma.conf.js',
+    browsers: ['PhantomJSCustom', 'Chrome'],
+    debug: true
+  }, done).start();
 });
-
-gulp.task('watch-tests', ['watch', 'karma']);
 
 gulp.task('default', ['watch']);


src/qtcore/qml/qml.js
```diff
@@ -139,6 +139,7 @@ function construct(meta) {
     return item;
 }
 
+
```
ChALkeR: unrelated

gulpfile.js
```diff
@@ -57,7 +57,9 @@ gulp.task('min-qt', ['qt'], function() {
 
 gulp.task('build', ['qt', 'min-qt']);
 
-gulp.task('test', ['build'], function(done) {
+gulp.task('build-dev', ['qt']);
```
ChALkeR: This is just an alias to `qt`?

gulpfile.js
```diff
@@ -57,7 +57,9 @@ gulp.task('min-qt', ['qt'], function() {
...
+
+gulp.task('test', ['build-dev'], function(done) {
```
ChALkeR: Hm. I undertand why you want this, but this shouldn't go under the name `test`.
`test` should be full tests, and those should probably include testing the minification process.

For example in #57 `min-qt` failed on two commits due to mistypes. Those commits would have failed later, but I am not sure that it would be true for all cases.

Note that `qt-min` stage is not what is slowing down the CI.

Perhaps this should be a separate stage? Like `test-fast`?

gulpfile.js
```diff
@@ -68,16 +70,20 @@ gulp.task('test', ['build'], function(done) {
...
   new karma.Server({
     configFile: __dirname + '/karma.conf.js'
   }, done).start();
 });
 
-gulp.task('watch', ['build'], function() {
-  gulp.watch(qtcoreSources, ['build']);
+gulp.task('debug-tests', ['build-dev'], function(done) {
+  new karma.Server({
+    configFile: __dirname + '/karma.debug.conf.js'
```
ChALkeR: I'm not 100% sure, but it should be possible to do this in a nicer way.
Try 
```javascript
{
  configFile: __dirname + '/karma.conf.js',
  browsers: ['PhantomJSCustom', 'Chrome'],
  debug: true
}

File: karma.debug.conf.js [hunk 5]
```diff
@@ -0,0 +1,8 @@
+
+module.exports = function(config) {
+  require("./karma.conf.js")(config)
+  config.browsers.push("Chrome")
+  config.set({
+    debug: true,
+  });
+};
```
ChALkeR: I hope this file won't be needed (se the comment above), though I'm not sure.

gulpfile.js
```diff
@@ -68,16 +70,20 @@ gulp.task('test', ['build'], function(done) {
   }).start();
 });
 
-gulp.task('karma', ['build'], function(done) {
+gulp.task('watch-tests', ['build-dev'], function(done) {
```
ChALkeR: Perhaps just `qt`.

ChALkeR:
While you are here — can you think of better names to `qt` and `min-qt`?

henrikrudstrom:
'build-dev' and 'build-min' (actually should maybe be 'build-release' ?)

had the idea that the minification was slow, dont know why. 
seems that karma handles its own rebuilding somehow, dont know how it does that.

henrikrudstrom:
actually, now qt.min took a while...

[00:59:00] Starting 'build-dev'...
[00:59:00] Finished 'build-dev' after 261 ms
[00:59:00] Starting 'build-min'...
[00:59:04] Finished 'build-min' after 3.8 s

ChALkeR:
@henrikrudstrom 

> actually, now qt.min took a while...

> [00:59:00] Starting 'build-dev'...
[00:59:00] Finished 'build-dev' after 261 ms
[00:59:00] Starting 'build-min'...
[00:59:04] Finished 'build-min' after 3.8 s

I agree that it is noticeable on the local machine, but on Travis a single run takes about 1½ minutes, around 40 seconds from which is the `npm install` stage, and 3 seconds are not a major change there. I would still prefer to have CI run the minification stage.

henrikrudstrom:
I changed it in the last commit, now the main `test` task does the full minification, but the watch tests only builds qt.js
what about the current task names, they ok?

gulpfile.js
```diff
@@ -55,7 +55,16 @@ gulp.task('min-qt', ['qt'], function() {
              .pipe(gulp.dest('./lib'));
 });
 
-gulp.task('build', ['qt', 'min-qt']);
+gulp.task('build', ['build-dev', 'build-rel']);
```
ChALkeR: `build` == `build-dev` and `build-rel`.
`build-rel` == `build-dev` and some actions.

This means that `build` == `build-rel`, so a separate `build-rel` entry isn't needed — just call it `build`.

gulpfile.js
```diff
@@ -45,7 +45,7 @@ gulp.task('qt', function() {
              .pipe(gulp.dest('./lib'));
 });
 
-gulp.task('min-qt', ['qt'], function() {
+gulp.task('build-rel', ['build-dev'], function() {
```
ChALkeR: s/build-rel/build/, see explanation in the previous comment.

karma.debug.conf.js
ChALkeR: This file is not needed now.

ChALkeR:
@henrikrudstrom The names look ok to me.
Take a look at comments on https://github.com/qmlweb/qmlweb/pull/65/files

ChALkeR:
@henrikrudstrom Could you squash the commits into a single one, please? You could do that in the same `git rebase` that you will use to rebase this against master.

LGTM if squashed.

ChALkeR:
@henrikrudstrom You forgot the Reviewed-By and PR-URL fields, I will amend and force-push.

ChALkeR:
Force-pushed.

Issue #51, opened 2016-02-19
Remove `global` usage


ChALkeR:
This should be probably done after merges.

Instead of using global, everything should be added to a single object (`qmlweb`? `QmlWeb`? `QMLWeb`? `QML`?) and that object should be added as a `window` property on browsers or directly exported in CJS.

This will allow us to ditch `encapsulate.begin.js` and `encapsulate.end.js` files which are actually not js files and make the build process simplier.

ChALkeR:
This should be also a step to fix #41.

ChALkeR:
`encapsulate.begin.js`/`encapsulate.end.js` were already removed, but the `global` is still used.
I'm keeping this open until that is fixed.

akreuzkamp:
I'm for `QmlWeb` this is common naming for namespaces. :)

ChALkeR:
@akreuzkamp See #285 =).

Issue #45, opened 2016-02-19
Consider using browserify or webpack


ChALkeR:
Atm, we bundle `uglify-js` is `src` dir and have our own encapsulate wrappers. That is affecting several things and has to be fixed in some way, and using browserify or webpack to make sure that we don't have to deal manually with various module systems would be nice.

ChALkeR:
This partially blocks #45 (I could use a work-around, but that would be ugly).

ChALkeR:
This also should be the way to fix #41.

henrikrudstrom:
my vote goes for Browserify, i understand Webpack doesnt play well with npm

Plaristote:
I've looked it up as well, and I lean towards Browserify as well.

aspidites:
We use web pack at work so I have a bit of experience with it. I'm not sure if browserify has equivalents, but I really like the following for webpack:
 - hot module reloading
 - Automatic bundling of common functionality between modules
- loaders for various precompilers
- dev server

ChALkeR:
I don't think that we need to over-complicate things, after all.
It should be good as it is now.

We should support those, though, but that's a separate matter.

ChALkeR:
I am going to close this.

I proposed this mainly for these reasons:
 * We were using uglify-js and needed a better way to load it.
 * I thought that it would help us with coverage analysis, by moving the build stage (#40, #66).
 * I thought that this would help us to split the parser (#41) and move it to a separate repo (#173).
 * I thought that this would help us to solve cross-element dependencies in an automatic way.

All of those are fixed now with simplier methods.

Feel free to reopen if you want to continue the discussion, though.

henrikrudstrom:
I would like to open this again. one thing that isnt addressed by the recent changes is using the qml engine in node.js environment, might be a bit of a niche application, but incidentally thats something that i need right now, so i'd be happy to do the work!

ChALkeR:
@henrikrudstrom Ah, fine, thanks for reopening!

Could you provide more details on the usecase? I was under an impression that only `parser` is needed in a non-browser enviroment now.

henrikrudstrom:
Im going to use the declarative engine to define solid 3d models, using a c++ library, so I need to be able to run it server side or in a nwjs application.

Plaristote:
True, we need to properly export our stuff so that it can be required from node.js.
We haven't really done that, have we ?

henrikrudstrom:
[...]

PR #66, opened 2016-02-22
Get source mappings to work on coverage

henrikrudstrom:
Coveralls cannot show the coverage reports because lib/qt.js is not in the repo. would be nice to have.
should be possible to use the sourcemaps somehow, but cant get the package karma-sourcemap-loader to work. 
Any idea why this does not work?

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -4,10 +4,10 @@ module.exports = function(config) {
     frameworks: ['jasmine'],
     files: [
       'lib/qt.js',
-      //{
-      //  pattern: 'lib/qt.js.map',
-      //  included: false
-      //},
+      {
+        pattern: 'lib/qt.js.map',
+        included: false
+      },
       'tests/common.js',
       'tests/failingTests.js',
       'tests/*/**/*.js',
@@ -24,7 +24,7 @@ module.exports = function(config) {
       dir: 'coverage/'
     },
     preprocessors: {
-      //'lib/qt.js.map': ['sourcemap'],
+      'lib/qt.js.map': ['sourcemap'],
       'lib/qt.js': ['coverage'],
 
     },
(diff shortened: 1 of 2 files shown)


karma.conf.js
```diff
@@ -4,11 +4,16 @@ module.exports = function(config) {
...
+      {
+        pattern: 'lib/qt.js.map',
+        included: false
+      },
       'tests/common.js',
       'tests/failingTests.js',
       'tests/*/**/*.js',
       { pattern: 'tests/*/**/*.qml', included: false },
-      { pattern: 'tests/*/**/*.png', included: false }
+      { pattern: 'tests/*/**/*.png', included: false },
```
ChALkeR: Unrelated

ChALkeR:
@henrikrudstrom Test coverage is calculated by [istanbul](https://www.npmjs.com/package/istanbul) by instrumenting the tested files before the actual tests are done. The test results are then stored locally and uploaded to Coveralls from Travis.

Atm, istanbul does not support reading sourcemaps when producing the coverage output. There is a tool called [lcov-sourcemap](https://www.npmjs.com/package/lcov-sourcemap) that should be able to transform the coverage output based on sourcemaps, but it looks not actively developed and not actively used. No idea if that works.

The proper solution to this would be to instrument `src/**/*.js`, not `lib/qt.js`, and build a special instrumented version for testing. Note that there are some files that are not actually proper js — for example, the encapsulate helpers.

I was planning to deal with this after we ditch the encapsulate helpers (blocked by #51), and, perhaps, after #45.

Both of those are actully blocked by the merging being completed first.

ChALkeR:
In the meantime, you can always view your coverage report (over a single file, regrettably) in the local `./coverage` directory =)

henrikrudstrom:
Guess we have to live with that for now.

ChALkeR:
I suppose that the issue behind this was fixed in #231.
Closing this as fixed.

If I am wrong and there was anything else that this PR was intended to achieve, feel free to reopen. =)

PR #63, opened 2016-02-21
Property tests

henrikrudstrom:
Added more tests:
- test basic property types
- test aliased property
- test bindings immediate update
- test reference (grand)parents by id

the last test tries to reference a parent by its Id, but it fails with an error on the parser level.

PR #67, opened 2016-02-22
Generated tests

henrikrudstrom:
Generates basic instantiation tests, can be used as template for further tests if the "//Generated" comment is removed. +5% coverage :)
depends on #63

### DIFF
--- a/tests/Initialize/runner.js
+++ b/tests/Initialize/runner.js
@@ -1,0 +1,79 @@
+describe('Initialize.loadQml', function () {
+  it('it can load qml without a file', function () {
+    var div = loadQml("import QtQuick 2.0\nItem {}\n");
+  });
+});
+
+var modules = {
+  "QtQuick": {
+    _version: "2.5",
+    _depends: {},
+    AnimatedImage: {},
+    Animation: {},
+    Behavior: {},
+    BorderImage: {},
+    Column: {dom: true},
+    DoubleValidator: {},
+    Flow: {},
+    Font: {},
+    FontLoader: {},
+    Grid: {dom: true},
+    Image: {dom: true},
+    IntValidator: {},
+    ListElement: {},
+    ListModel: {},
+    ListView: {dom: true},
+    Loader: {},
+    MouseArea: {dom: true},
+    NumberAnimation: {},
+    ParallelAnimation: {},
+    PropertyAnimation: {},
+    Rectangle: {dom: true},
+    RegExpValidator: {},
+    Repeater: {dom: true},
+    Rotation: {},
+    Row: {dom: true},
+    Scale: {},
+    SequentialAnimation: {},
+    State: {},
+    SystemPalette: {},
+    Text: {dom: true},
+    TextInput: {dom: true},
+    Timer: {},
+    Transition: {},
+    Translate: {}
+  }
+};
+
+function testModule(modules, testFunc) {
+  Object.keys(modules).forEach(function (module) {
+    var imports = "import " + module + " " + modules[module]._version + "\n";
+    for (var i in module._depends) {
+      imports += "import " + module._depends[i] + "\n";
+    }
+    Object.keys(modules[module]).forEach(function (element) {
+      if (element[0] === ("_")) return;
+      testFunc(module, element, imports, modules[module][element]);
+    });
+  });
+}
+
+testModule(modules, function (module, element, imports, options) {
+  describe('Initialize.' + module, function () {
+    it(element, function () {
+      var src = imports + element + " { }\n";
+      var engine = loadQml(src);
+      var div = engine.rootElement;
+      var qml = engine.rootObject;
+      expect(div).not.toBe(undefined);
+      if (options.dom) {
+        expect(div.className).toBe(element);
+        expect(div.qml).not.toBe(undefined);
+        expect(div.qml.Component).not.toBe(undefined);
+      } else {
+        expect(qml.Component).not.toBe(undefined);
+      }
+      div.remove();
+    });
+  });
+});
--- a/tests/QMLEngine/qml/ScopeLeaf.qml
+++ b/tests/QMLEngine/qml/ScopeLeaf.qml
@@ -2,5 +2,5 @@ Item {
   property int value: 1
   property int rootValue: rootItem.value
   property int parentValue: parentItem.value
-  property int totalValue: rootValue + parentValue
+  property int totalValue: (rootValue + parentValue) * value
 }
--- a/tests/failingTests.js
+++ b/tests/failingTests.js
@@ -12,5 +12,14 @@ window.failingTests = {
     imports: [
       'Javascript'
     ]
+  },
+  Initialize: {
+    QtQuick: [
+      'Translate',
+      'Scale',
+      'Rotation',
+      'Font',
+      'AnimatedImage'
+    ]
   }
 };
--- a/tests/failingTests.js
+++ b/tests/failingTests.js
@@ -11,9 +11,6 @@ window.failingTests = {
   QMLEngine: {
     imports: [
       'Javascript'
-    ],
-    scope: [
-      'can reference parent items id'
     ]
   }
 };
--- a/tests/common.js
+++ b/tests/common.js
@@ -13,6 +13,15 @@ function prefixedQmlLoader(prefix) {
   }
 }
 
+function loadQml(src, opts) {
+  var div = document.createElement('div');
+  var qml = new QMLEngine(div, opts || {});
+  qml.loadQML(src);
+  qml.start();
+  document.body.appendChild(div);
+  return qml;
+}
+
 (function() {
   var describeOrig = describe;
   var itOrig = it;


ChALkeR:
Hm. Several considerations here:

 1. To achieve the same effect, a lot of empty files is not needed. You could use `loadQML` and directly pass the content to it, without creating all those files.
 2. There is also no need to pre-generate them. You can just write a test, for example `Initialization/runner.js`, that would store an `Array` of tested elements, or a an `Array` of modules, like:
 
 ```javascript
 var tests = {
    'QtQuick 2.5': [
      'Rectangle', …
    ],
    …
 }
 ```
  and then just call `it` for each of them. Note that this way you could add even tests for unimplemented elements, but blacklist them in the failingTests — then it would be obvious which elements are missing.

 3. Are all of those supposed to be created in Qt QML? i.e.: does `qmlscene` work without warnings on all those files? Most probably yes, but it should be better to re-check.
 
 4. Note the code coverage measure is not absolute — these tests are valuable, but they don't verify that the elements actually work as they are supposed to, so there should be also behaviour/render tests for those. In separate PRs, of course.

henrikrudstrom:
is it possible to load inline qml somehow, (tried but the code is quite convoluted)... if so we could skip generating files for the basic tests and just do a foreach on each `describe()` like you do in the render tests...

henrikrudstrom:
@ChALkeR 
1. thats what i want, tried briefly, but doesnt it need a path to resolve the imports properly?
2. +1
3. i tried a bit with the `qml` command (not sure what the difference is with qmlscene), havent figured out how to do the screenshot, was hoping it was doable in pure qml (take a QtGraphicalEffect MaskImage or something and save the buffer to a file, but not sure that works without c++). 
4. I know, was just a start to get some basic coverage... but i want to see if we can add some more tests that apply to all components ("className") etc...

henrikrudstrom:
@ChALkeR I tried to use the loadQml again, i get a `JS_Parse_Error`.
ill add some comments in the commited code, if you have a chance to look at it.

Another issue is that Karma seems to compress the qt.js source for some reason, all line numbers are refering to line 9, and if you rund `test-debug` and inspect the code, then qt.js has been mangled...

ChALkeR:
> if you rund test-debug and inspect the code, then qt.js has been mangled...

Try adding `preprocessors: {}` or `preprocessors: {'lib/qt.js': []}` to the debug karma configuration.

henrikrudstrom:
@ChALkeR addint preprocessors worked for the stack trace

the parser gives this error message: `'Unexpected token name undefined, expected punc {'`
the source code i pass to loadQml is `Import QtQuick 2.0\n\rItem{}\n\r` ive tried different variation with more or less spaces, newlines and ;... 
any ideas?

ChALkeR:
```javascript
    var div = document.createElement('div');
    var qml = new QMLEngine(div, {});
    qml.loadQML('import QtQuick 2.0\nItem {}\n');
    qml.start();
    div.remove();
```

works for me on current master.

henrikrudstrom:
seems to work now, thanks

henrikrudstrom:
just found there is also `qmlviewer' 
it has these options, didnt get time to get it to work, but looks promising
```
-recordfile <output> ..................... set video recording file
                                              - ImageMagick 'convert' for GIF)
                                              - png file for raw frames
                                              - 'ffmpeg' for other formats
  -recorddither ordered|threshold|floyd .... set GIF dither recording mode
  -recordrate <fps> ........................ set recording frame rate
  -record arg .............................. add a recording process argument
  -autorecord [from-]<tomilliseconds> ...... set recording to start and stop
```

ChALkeR:
@henrikrudstrom `qmlviewer` is for old QML from Qt 4.x. It's not compatible with `QtQuick 2.x`.

tests/Initialize/runner.js
```diff
@@ -0,0 +1,70 @@
+describe('Initialization.loadQml', function() {
```
ChALkeR: `Initialize` is used below. s/Initialization/Initialize/ for consistency.

tests/QMLEngine/qml/PropertiesAlias.qml
```diff
@@ -5,4 +5,4 @@ Item {
     id: child
     x: 125
   }
-}
+}
```
ChALkeR: unrelated

tests/QMLEngine/qml/ScopeLeaf.qml
```diff
@@ -2,5 +2,5 @@ Item {
   property int value: 1
   property int rootValue: rootItem.value
   property int parentValue: parentItem.value
-  property int totalValue: rootValue + parentValue
+  property int totalValue: (rootValue + parentValue) * value
```
ChALkeR: unrelated?

tests/QMLEngine/qml/ScopeMid.qml
```diff
@@ -1,15 +1,15 @@
...
+    }
 
-  ScopeLeaf {
-    id: childB
-    value: 4
-  }
+    ScopeLeaf {
+        id: childB
+        value: 4
+    }
```
ChALkeR: unrelated

tests/failingTests.js
```diff
@@ -1,7 +1,7 @@
 window.failingTests = {
   Render: {
     Async: [
-      'NumberAnimationAutorun'
+      'NumberAnimationAutorun', 'Image'
```
ChALkeR: Um. Did it start failing? It's not failing for me, did something happen? What's the error?

tests/Initialize/runner.js
```diff
@@ -0,0 +1,70 @@
...
+    {name: 'Scale', tags: []},
+    {name: 'SequentialAnimation', tags: []},
+    {name: 'State', tags: []},
+    {name: 'SystemPalette', tags: []},
+    {name: 'Text', tags: ["dom"]},
+    {name: 'TextInput', tags: ["dom"]},
+    {name: 'Timer', tags: []},
+    {name: 'Transition', tags: []},
+    {name: 'Translate', tags: []},
+]
```
ChALkeR: This does not support modules other than QtQuick. Could you add other modules support, too?
Also, this is a bit vebrose, `module.element.options` would be more compact.

Like this:
```javascript
var modules = [
  `QtQuick 2.5': {
    AnimatedImage: {},
    Animation: {},
    Behavior: {},
    BorderImage: {},
    Column: {dom: true},
    …
 }
};
```

tests/failingTests.js
```diff
@@ -12,8 +12,12 @@ window.failingTests = {
...
-    scope: [
-      'can reference parent items id'
-    ]
+  },
+  Initialize: {
+    Translate: ["can be loaded"],
+    Scale: ["can be loaded"],
+    Rotation: ["can be loaded"],
+    Font: ["can be loaded"],
+    AnimatedImage: ["can be loaded"],
```
ChALkeR: I don't like how exactly this does look.
If you we use `module.element.options`, could we use `Initialize.module.element` as the test id?
Then, this list will look like this:
```javascript
Initialize: {
  'QtQuick 2.5': [
    'Translate',
    'Scale',
    …
  ]
}

**Comment by henrikrudstrom** [comment]
did a rebase, will look into your suggestions tommorow

**Comment by labsin** [comment]
To export to an image, you could use
http://doc.qt.io/qt-5/qquickitem.html#grabToImage
I could write a little command-line script and batch as many examples as needed. The source of the qml command is not that big either.

**Comment by henrikrudstrom** [comment]
that would be awesome, would be really cool if all you need for a test case is a qml file!
is that one available from qml or do we need to compile something in c++?
Think we should make the script as simple as possible and control batches etc from node.js, unless there is a signficant performance overhead.

**Comment by labsin** [comment]
What I found needs to be compiled. I might give out a shot this evening. I would create the images from qt it self before the tests and not while, because it will take some seconds per qml file.

**Comment by henrikrudstrom** [comment]
sounds good. We could run it with a gulp task. Only question is if we want to have a c++/qt project in the main repo? @ChALkeR @Plaristote ?

**Comment by ChALkeR** [comment]
@henrikrudstrom @labsin I don't think that we should bring c++/qt stuff to the main repo. It's not needed for tests — just png images here would be fine enough.

What about creating a separate repo that would have a tiny c++/qt qml render lib and a js wrapper around it?

**Comment by henrikrudstrom** [comment]
changed the test list as you suggested @ChALkeR, if you have anything you want tweaked, feel free to push to this commit, wont be able to work on this anymore today.

**Comment by ChALkeR** [comment]
@henrikrudstrom Ok, thanks! I will push my changes here today (in several hours).

**Comment by labsin** [comment]
I forked the qml tool from qtdeclarative and made it a screenshot tool:
https://github.com/labsin/qmlscreenshot
Should be strait forward to build.
You could use something like: 'find -name *.qml -exec ./qmlscreenshot -o output {} +'  to save all the qml files and put them in ./output in the same directory structure.
Feel free to fork and propose additions.

**Comment by ChALkeR** [comment]
Rebased against `master`.

**Comment by ChALkeR** [comment]
Changed order and squashed two commits into one.

File: tests/QMLEngine/qml/ScopeLeaf.qml (hunk 3 above)
Comment (ChALkeR) [inline tests/QMLEngine/qml/ScopeLeaf.qml]: Actually, this is from a commit that fixes the scopes test. It's does not belong to this PR, but it (the commit that fixes the scopes tests) is LGTM nevertheless.

**Comment by ChALkeR** [comment]
@henrikrudstrom Could you please check my edits? If those look good, I will squash them into your commit to keep this clean. Would that be ok?

File: src/qtcore/qml/lib/parser.js [hunk 8]
```diff
@@ -1725,4 +1725,5 @@ function parseQML(src) {
 
 if (typeof global != "undefined") {
   global.qmlparse = qmlparse;
+  global.convertToEngine = convertToEngine;
```
ChALkeR: Could you explain why we need this change, please?

henrikrudstrom: we dont need that, thought i removed it the last commit

henrikrudstrom:
Looks good to me. Squash it!

ChALkeR:
Squashed, LGTM.

ChALkeR:
Ow, sorry. Rebasing.

ChALkeR:
Rebased.

PR #70, opened 2016-02-23
remove code compression in karma

henrikrudstrom:
karma currently compresses the code under test, makes for bad stack traces...

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -84,6 +84,8 @@ gulp.task('test-debug', ['watch-dev'], function(done) {
   new karma.Server({
     configFile: __dirname + '/karma.conf.js',
     browsers: ['PhantomJSCustom', 'Chrome'],
+    preprocessors: {},
+    reporters: ['progress'],
     debug: true
   }, done).start();
 });


ChALkeR:
I don't think this works, you have to override that option only in `test-debug` (like the other karma options).

henrikrudstrom:
@ChALkeR works here, but i can add it to the gulp task if you want, but then it should be added to test-watch as well, good to have a correct stacktrace there as well...

ChALkeR:
@henrikrudstrom It broke the coverage reports, see the `[show all checks]` at this PR.

henrikrudstrom:
thats a good argument... will update soon

karma.conf.js
```diff
@@ -19,7 +19,7 @@ module.exports = function(config) {
       dir: 'coverage/'
     },
     preprocessors: {
-      'lib/qt.js': ['coverage']
+      'lib/qt.js': ['coverage'],
```
ChALkeR: unwanted edit

gulpfile.js
```diff
@@ -76,7 +76,10 @@ gulp.task('test', ['build'], function(done) {
 
 gulp.task('test-watch', ['watch-dev'], function(done) {
   new karma.Server({
-    configFile: __dirname + '/karma.conf.js'
+    configFile: __dirname + '/karma.conf.js',
+    preprocessors: {
+      'lib/qt.js': []
+    },
```
ChALkeR: Do we want to disable code coverage generation in test-watch?

henrikrudstrom:
good point, hopefully we can get both at the same time when we switch to browserify/webpack
@ChALkeR should be good to go now

gulpfile.js
```diff
@@ -84,6 +84,9 @@ gulp.task('test-debug', ['watch-dev'], function(done) {
   new karma.Server({
     configFile: __dirname + '/karma.conf.js',
     browsers: ['PhantomJSCustom', 'Chrome'],
+    preprocessors: {
+      'lib/qt.js': []
+    },
```
ChALkeR: I just tested it, just `preprocessors: {},` also works. No need to specify the file.
Also, `reporters: ['progress']` would remove the generation of an empty coverage report.

ChALkeR:
LGTM, merging.

ChALkeR:
I'm squashing this.

PR #72, opened 2016-02-24
Border fix for Rectangle

igosoft:
Default color of border is black now. Default border width is 1px when color of border has been been specified but not the width.


henrikrudstrom:
lgtm

ChALkeR:
Two questions:
 1. Do we have `box-sizing: border-box` enabled? If no, this would make the border _around_ the rectagle, not _inside_, as it should be. I can't check atm, sorry.
 2. Could we have render a test for this? Here is an (approximate) source:
 
 ```qml
  import QtQuick 2.0
 
  Rectangle {
    width: 20
    height: 20
    border.width: 5
    Rectangle {
      width: 5
      height: 5
    }
    Rectangle {
      width: 5
      height: 5
      x: 5
      border.color: 'red'
    }
    Rectangle {
      width: 5
      height: 5
      x: 7
      y: 7
      border.width: 2
    }
  }
 ```

henrikrudstrom:
Added a render test, it fails so i suppose the border is added on the outside... 
tried to run the test in chrome to see the actual image generated, but the render tests dont seem to run in the browser, can anyone confirm that?

akreuzkamp:
lgtm! :)

@ChALkeR: No, we don't have that, yet. I didn't know of that property back when I implemented borders, but I remember that border-rendering caused me head-aches because of that . We definitely want box-sizing: border-box (assuming it's available everywhere) and finally get correct border rendering :)

akreuzkamp:
iirc that if width==0 ? "1px" thingie is one of these box-sizing workarounds ;)

igosoft:
@akreuzkamp It's just the use case when user doesn't set a border width but a color only.

ChALkeR:
Rebased.

ChALkeR:
Hm. Placing the `box-sizing` aside, this looks like it doesn't set the 1px border on this example:
```
  Rectangle {
    width: 10
    height: 10
    border.color: 'red'
  }
```

But it should be.

ChALkeR:
Looks like the `borderStyle` is set to `"none"`.

src/qtcore/qml/elements/QtQuick/Rectangle.js
```diff
@@ -19,6 +19,7 @@ registerQmlType({
         this.dom.style.borderRadius = newVal + "px";
     });
     this.border.colorChanged.connect(this, function(newVal) {
+        this.dom.style.borderWidth = this.border.width == 0 ? "1px" : this.border.width + "px";
         this.dom.style.borderColor = newVal;
         this.dom.style.borderStyle = this.border.width == 0 || newVal == "transparent"
```
ChALkeR: On this line, even when `this.dom.style.borderWidth` is set to `"1px"`, when `this.border.width == 0`, there is no border because of the `this.dom.style.borderStyle` being set to `"none"`.

igosoft: Thanks for the review.  (fix committed)

ChALkeR:
Just checked how does Qt behave.

 1. If not defined, `border.color` is always `#000000`, it does not depend on `border.width` being set or not.
 2. If not defined, `border.width` is always `1`, it does not depend on `border.color` being set or not.
 3. If neitherer of `border.color` and `border.width` are explicitly defined, then the border is invisible for some reason.
 4. If either of them is defined, border becomes visible according to the actual `border.color`/`border.width` values.

This gives us a following algorithm:
 1. Set `border.color` to `#000000` and `border.width` to `1` before registering `border.colorChanged `border.widthChanged` listeners.
 2. Don't set the actual DOM border width/color outside of those listeners.
 3. Set both DOM border width and color in each of those listeners according to `border.color` and `border.width`.

Plaristote:
> Do we have box-sizing: border-box enabled? If no, would this make the border around the rectangle, not inside, as it should be? I can't check atm, sorry.

> Edit: actually, if this problem exists, then it was there even before this PR, so this should not stop this PR from landing.

@ChALkeR: actually I've met this problem while fixing the filling issue with BorderImages. We're not using the proper box-sizing.
I've seen one of the repository we're going to merge creates a `globalStylesheet` object... so I don't know if this problem should be fixed now, or once the `globalStylesheet` is merged.

I'm not even sure using `globalStylesheet` is a good idea (apart from things like `box-sizing`, is it really more efficient than directly settings the styles on the elements) ? I don't know.

About the question for the availability of `box-sizing`, it is available everywhere. It has a long history, related to IE: https://en.wikipedia.org/wiki/Internet_Explorer_box_model_bug

ChALkeR:
@Plaristote This commit adds `this.dom.style.boxSizing = "border-box";` to `Rectangle` now. This should fix the size of the Rectangles, but not the inner elements placement — those are misplaced. Perhaps that should be fixed in a follow-up PR.

About the `globalStylesheet` — let's just set the style per-items for now, furhter optimization/cleanup could be discussed later, I think.

igosoft:
Do we miss anything else here ?

ChALkeR:
@igosoft See https://github.com/qmlweb/qmlweb/pull/72#issuecomment-190023795

ChALkeR:
Atm this doesn't match with the Qt behaviour, and I suppose that this could be easily fixed.

igosoft:
I found two new issues related to Rectangle rendering
1) border width changes size of Rectangle when  border.width > width / 2
2) Item width/height set to zero must to hide div. Otherwise div border will be rendered (the fix #105  )

src/qtcore/qml/elements/QtQuick/Rectangle.js
```diff
@@ -4,36 +4,39 @@ registerQmlType({
...
+        this.css.borderRadius = newVal + "px";
     });
     this.border.colorChanged.connect(this, function(newVal) {
-        this.dom.style.borderColor = newVal;
-        this.dom.style.borderStyle = this.border.width == 0 || newVal == "transparent"
-                                            ? "none" : "solid";
+        this.css.borderColor = newVal;
+        // if no border width was set then take default value (1px)
+        if (this.css.borderWidth == "0px") {
+            this.css.borderWidth = "1px";
```
ChALkeR: Incorrect.

Testcase:
```qml
import QtQuick 2.0

Rectangle {
  width: 200
  height: 200
  border.width: 0
  border.color: 'red'
}
```

This should help:
```diff
-        if (this.css.borderWidth == "0px") {
-            this.css.borderWidth = "1px";
-        }
+        this.css.borderWidth = this.border.width + "px";
```

ChALkeR:
> 1) border width can change size of Rectangle when border.width > width / 2
> 2) Item width/height set to zero must to hide div. Otherwise div border will be rendered (the fix #105 )

This could be fixed in a separate PR, I guess =). Could you apply the above fix I mentioned, so I could merge this?

ChALkeR:
@henrikrudstrom Btw, I will reduce the testcase width when landing, it's unnecessary.
It's better to keep the image smaller, the testcases pass a little bit faster that way.

igosoft:
@ChALkeR I added #107 to trace the border size issue. #72 looks like it's ready to be merged...

ChALkeR:
@igosoft I will add a default `this.css.borderColor = 'black'` it it, ok? I think it's missing here.

igosoft:
Indeed its good idea.

ChALkeR:
@igosoft Sorry, I'm removing the border-radius test from here. Rounded borders could be approximated and antialiased differently in browsers/qml, and those minor differences would break the test which expects pixel-perfect match. And that's not a bug, I think.

If you want to test border-radius not increasing the actual height, this could be done in some other way (i.e. a DOM property test).

igosoft:
@ChALkeR OK. Indeed border radius is a difficult subject. It should be treated separately.

ChALkeR:
@henrikrudstrom I squashed all the test updates here into your commit and split it into two files (the passing one and the failing one). Hope you don't mind =).

igosoft:
@ChALkeR we need an extra test for Rectangle. That rect must has width and/or height equal zero and border.width greater than 0.

ChALkeR:
@igosoft I already merged this, sorry. Could you file a separate PR to alter the tests or to create a new one?

PR #77, opened 2016-02-26
Fix parsing of properies named signal.

### LINKED ISSUE #76: cannot parse a property named 'signal'
Some standard element from QML have properties named 'signal' (QtQml.StateMachine.SignalTransition for instance).
However, the qml parser goes nuts when he meets the signal keyword used as a property name. We'll have to fix that at some point.

akreuzkamp:
So far if you wrote "signal:" as a property name the parsing would
fail because it handle signal as a keyword. Now this case is handled
that if signal is followed by a colon it's ok and handled as a
property.

This is completely untested, as I don't have a lot of time, these days. Please be so kind and add a auto test for me. :)

Fixes #76.

### DIFF
--- a/src/qtcore/qml/lib/parser.js
+++ b/src/qtcore/qml/lib/parser.js
@@ -1339,7 +1339,6 @@ function qmlparse($TEXT, exigent_mode, embed_tokens) {
         }
 
         function qmlsignaldef() {
-            next();
             var name = S.token.value;
             next();
             var args = [];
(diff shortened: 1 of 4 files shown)


henrikrudstrom:
Seems to work

akreuzkamp:
Thanks a lot for adding a test. Unfortunately it doesn't cover the change of this PR. This PR only covers the case, that you don't have a 'property' before the 'signal'. It's good to have that test, but this one  should also work without this patch. We need to test the case, where we have a line like 'signal: 10' only

henrikrudstrom:
I see, now it fails!
(also the Image test for some reason)

ChALkeR:
@henrikrudstrom The Image test is fixed in #80 (see also #79 for some actual Image fixes).

ChALkeR:
@henrikrudstrom Btw, you have typed `QtCuick` istead of `QtQuick`. I will fix that and squash.

ChALkeR:
Rebased, squashed the tests, fixed the tests formatting, added `div.remove()`.
@akreuzkamp As I understand, this should have passed now, but it's not. Am I correct?

ChALkeR:
Hm. I see. There was also an error in the testcase that I haven't noticed.
Thanks, @akreuzkamp!

LGTM.

ChALkeR:
@akreuzkamp This actually looks broken, see 9d3d8d34b68a44f2c1d8cb1230883bea1df8ecae

ChALkeR:
@akreuzkamp This actually looks broken, see 51813575b1d8107dce9ce5a91b864bf07c65ad2d

Issue #75, opened 2016-02-25
Render/Async/Image test is sometimes failing.


ChALkeR:
Noticed that in a CI job.
Will take a look at that.

ChALkeR:
Ok, the image is sometimes not shown (i.e. the render result is transparent).

ChALkeR:
It has the correct size, though. And if I add a rectangle — it is shown.
Looks like even after `status` is set to `Image.Ready`, it's not actually shown yet in some cases.

ChALkeR:
It looks like the actual DOM Image is loaded, but not yet shown when its `onload` event fires.

PR #78, opened 2016-02-26
adds an options parameter to createSimpleProperty

Plaristote:
The comment above `createSimpleProperty` spoke of an `options` parameter, which this commit now implements.

If the `options` parameter isn't an object, it'll be used as the default value. If it is an object, we look for the `readOnly` and `default` attributes to set the access policy and the default value.

This will fix the case(s?) when people have used the fourth parameter of `createSimpleProperty` to set default values (while this parameter was actually used to define the access policy).

### DIFF
--- a/src/qtcore/qml/elements/QtQuick/SystemPalette.js
+++ b/src/qtcore/qml/elements/QtQuick/SystemPalette.js
@@ -26,7 +26,7 @@ registerQmlType({
     var platform = 'OSX';
 
     for (var i = 0 ; i < attrs.length ; ++i)
-      createSimpleProperty("color", this, attrs[i], "ro");
+      createSimpleProperty("color", this, attrs[i], { readOnly: true });
     createSimpleProperty("enum", this, "colorGroup");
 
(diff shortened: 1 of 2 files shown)


henrikrudstrom:
LGTM

ChALkeR:
Travis failure is unrelated, it's Render.Async.Image test being flaky, see #75.

ChALkeR:
> This will fix the case(s?) when people have used the fourth parameter of createSimpleProperty to set default values (while this parameter was actually used to define the access policy).

Could you provide an example of those? If yes, than LGTM, if no — I would prefer not to introduce polymorphism here and just let everyone use `{default: 'whatever'}` as the fourth parameter.

ChALkeR:
Ah, actually it's in the https://github.com/qmlweb/qmlweb/commit/8c6ec044d51cdd8cc19bd7e521616149bc4f6f09 (which will not be merged as-is).

Ok, LGTM. We can alter that later (if we want) in time to 1.0.0 after everything would be merged.

PR #79, opened 2016-02-27
Image: set the status after dimensions

ChALkeR:
This is how Qt does things, it also gives us the correct dimensions inside the onStatusChanged callback.

The Image render tests has been modified to also check that the dimensions have been set correctly (it does not force them anymore).

/cc @Plaristote @akreuzkamp @henrikrudstrom


ChALkeR:
This was found during #75 investigation, but it is not the cause and does not solve #75.

henrikrudstrom:
LGTM

PR #82, opened 2016-02-27
tests: Serve qt.js.map from karma.

ChALkeR:
This will give us the original sources and lines in the browser developer tools in the debug mode.

/cc @akreuzkamp @henrikrudstrom

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -4,6 +4,7 @@ module.exports = function(config) {
     frameworks: ['jasmine'],
     files: [
       'lib/qt.js',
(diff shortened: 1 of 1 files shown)


henrikrudstrom:
@ChALkeR, tried running gulp test-debug, it doesnt generatate any coverage files at all, the other test tasks just gives me coverage for qt.js only...

ChALkeR:
@henrikrudstrom This is not about the coverage, it's about the source maps in the browser console window.

henrikrudstrom:
ah, thats also great! works and LGTM

Issue #46, opened 2016-02-19
Merge and commit landing process


ChALkeR:
### How the merge process goes.

Key points:
1. The commits should have `PR-URL`, `Reviewed-By`, and (sometimes) `Fixes` fields.
2. `PR-URL` and `Fixes` should be full links to relevant GitHub PR and Issues accordingly.
3. Fast-forward is preferred instead of merges to reduce commit noise. **Edit:** I'm not sure about this, especially for the PRs with many commits, and perhaps we should discuss this =).
4. This is lower priority, but if possible — we should get nice badges at the PRs to make it clear that those were merged.
5. The Travis CI checks should pass.
6. AUTHORS should contain a full list of all contributors.

How to achieve this:
1. Rebase the commits that you are merging against current master.
2. Add header fields (`PR-URL`, `Reviewed-By`, `Fixes`) to them.
   
   This and step 1 could be done together, using `git rebase -i master` and marking all commits with an `e` to amend them.
3. If possibble — force-push to the branch that is being merged, this way the PR will be marked as `Merged` afterwards.
4. Wait around two minutes for Travis CI tests to complete and make sure that those are green.
   
   _Note to self:_ I was not waiting for them at this point, and this is why we got red Travis statuses on the closed pull requests, because the merge itself failed.
5. Land the commits on `master` and push.

/cc @qmlweb/collaborators

ChALkeR:
GitHub is not very good at this, afaik the «rebase» button is one of the most wanted features here.

henrikrudstrom:
found this: https://github.com/git-land/git-land 
regarding 3. found this in their docs:

> If merging a PR, append [close #<PR number>] to the last commit message so that Github will close the pull request when the merged commits are pushed.

ChALkeR:
A status update: GitHub added «squash and merge» option, so just using that for indivudual commits is fine, as long as you add the `PR-URL` and `Reviewed-By` information at the merge time (which could be added thought GitHub web ui).

I used that for merging #250, for example.

ChALkeR:
I guess I will just close this, enforcing the hard rules here just adds churn, «Rebase and merge» is now good enough. Will think about it more if there would be any issues with that.

ChALkeR:
[...]

PR #83, opened 2016-02-28
fix BorderImage not filling the middle region

Plaristote:
The current implementation for BorderImages doesn't fill the middle region as described in http://doc.qt.io/qt-5/qml-qtquick-borderimage.html
Though we can't chose between repeat and scaled mode, this change allows our BorderImage object to render a little bit more like the original implementation.

### DIFF
--- a/src/qtcore/qml/elements/QtQuick/BorderImage.js
+++ b/src/qtcore/qml/elements/QtQuick/BorderImage.js
@@ -52,7 +52,8 @@ registerQmlType({
         this.dom.style.MozBorderImageSlice = this.border.top + " "
                                                 + this.border.right + " "
                                                 + this.border.bottom + " "
-                                                + this.border.left;
+                                                + this.border.left + " "
+                                                + "fill";
         this.dom.style.MozBorderImageRepeat = this.horizontalTileMode + " "
                                                     + this.verticalTileMode;
         this.dom.style.MozBorderImageWidth = this.border.top + " "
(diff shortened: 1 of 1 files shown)


ChALkeR:
LGTM.

ChALkeR:
Relevant css doc: https://developer.mozilla.org/en-US/docs/Web/CSS/border-image-slice.

> Though we can't chose between repeat and scaled mode, this change allows our BorderImage object to render a little bit more like the original implementation.

What's wrong with the repeat/scaled mode?

QML doc states:
> The middle (region 5) is scaled according to both horizontalTileMode and verticalTileMode.

CSS doc states:
> `fill` is a keyword whose presence forces the use of the middle image slice to be displayed over the background image, its size and height are resized like those of the top and left image slices, respectively.

That _should_ play together nicely, doesn't it?

Plaristote:
Didn't you write something about the merging process ?
I can't figure out how to make these pull request work without creating a merge commit.

Even when I rebase the branch on master before doing the merge, there's a merge commit. So I tried to rebase this onto master instead, and now there's no merge commit, but this pull request isn't considered as merged.

ChALkeR:
@Plaristote Actually, those are not final and I'm open for discussion, perhaps merge commits are useful for large merges.

The way how I was doing it is documented in #46, it's basically the following:
 1. Rebase the branch against master and force-push.
 
 `git rebase -i` allows to amend all the commits to set relevant fields (PR-URL/Reviewed-By) in all of themn.
 
 Don't forget to update the local master branch to make sure that it's synced with the remote master branch before rebasing.
 2. Wait until Travis finishes the test build and updates the status of the PR.
 3. Manually merge to master and push.
  
  `git checkout master; git merge yourbranch` — this will do a fast-forward if the branch is rebased against master, then `git push`.
 
 If it doesn't do a fast-forward — something was wrong in the first step. You could always use the `--ff-only` option of `git merge` to double-check that, but I generally don't.

Issue #81, opened 2016-02-27
internal dev project


henrikrudstrom:
How about we make a separate project with all files and tools relevant to development of qmlweb itself but would clutter the official distribution: 
* screenshot utility + gulp task
* design documents
* .qmlproject file for the test files etc.

we can add it as a submodule in the main repo, so it would not be downloaded automatically but still we can reference the path consistently in gulp tasks etc.

ChALkeR:
> screenshot utility + gulp task

Not sure. A screenshot utility would be best fit as a Node.js module that uses Qt to create screenshots for qml files. It's currently located at https://github.com/qmlweb/qmlweb-dev, but I believe that it should be improved before getting here.

We could add a separate gulp task that runs all the tests using that module, creating the missing screenshots, if necessary.

> design documents

Should be placed either here or in the wiki, most likely. Note that the wiki is already a git repo.

> .qmlproject file for the test files etc.

We could just add a qmlproject file for test files here, to the `tests` dir =). Not sure why that needs to be put in a separate repo.

PR #84, opened 2016-02-29
test rendering/devtools project 

henrikrudstrom:
i created two new projects:
[qmlbook/shorty](https://github.com/qmlweb/shorty/) a fork of qmlbook/shorty which is a simple screenshot utility. I modified it so you can specify a loader qml (which handles the actual screenshot creation) and a test qml to be rendered. (we should probably rename or merge it into qmlweb-dev)

[qmlweb-dev](https://github.com/qmlweb/qmlweb-dev) is a project that can be cloned into the qmlweb main project. we can put optional dev stuff here. for now it contains the screenshot renderer.

just a proof of concept for now, but would like to hear your thoughts. 
### Clone and build:

in your qmlweb folder:
`clone https://github.com/qmlweb/qmlweb-dev.git dev`

`cd dev`

`./build.sh`
### Render tests:

`node dev/render.js`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ build/
 *.kdev4
 *kate-swp
 .idea
+dev
+*.autosave
--- a/qmlweb-dev-install.bat
+++ b/qmlweb-dev-install.bat
@@ -1,0 +1,4 @@
+git clone https://github.com/qmlweb/qmlweb-dev.git dev
+cd dev
+build.bat
+cd ..
--- a/qmlweb-dev-install.sh
+++ b/qmlweb-dev-install.sh
@@ -1,0 +1,4 @@
+git clone https://github.com/qmlweb/qmlweb-dev.git dev
+cd dev
+./build.sh
+cd ..


.gitignore
```diff
@@ -11,3 +11,5 @@ build/
 *.kdev4
 *kate-swp
 .idea
+dev
+*.autosave
```
ChALkeR: Add a newline at the end of file.

tests/Render/runner.js
```diff
@@ -34,6 +34,8 @@
     var ctx = canvas.getContext('2d');
     canvas.height = img.height;
     canvas.width = img.width;
+    ctx.fillStyle = "rgb(255,255,255)";
+    ctx.fillRect (0,0, canvas.width,canvas.width);
```
ChALkeR: `canvas.width

ChALkeR: `canvas.width` two times? Also, code style.
Should be `ctx.fillRect(0, 0, canvas.width, canvas.height);`.

ChALkeR:
This in fact does two things («integrate with qmlweb-dev» does not explain it good enough).

 1. Adds two more entries to `.gitignore` — this one looks unambiguous, LGTM on that.
 2. When comparing images in render tests, drops the opacity and compares them on the white background. Are we sure that we want that? I am viewing the transparency of the window as a _feature_ that should be tested. If some tests do not want it — they could always paint the background using a rectangle.

ChALkeR:
Could you split that into two separate commits, btw?

henrikrudstrom:
will split.
The reason is that i changed the background was to match the output of the screenshot utility, could find out how to make that transparent, but afaik its not possible to make the browser window transparent, so everything would be renderend agains a background...

henrikrudstrom:
rebased and moved screenshot background to separate commit
@ChALkeR @pavelvasev, @Plaristote LGTY?

Plaristote:
Shouldn't we use git submodules instead of a script ?

ChALkeR:
@Plaristote Yes, this should definitely use git submodules or a devdependency via npm — that depends on the purpose of the tools.

Note: I have not reviewed that repo yet, perhaps that should be done first before pulling this in here as a dependency.

ChALkeR:
@Plaristote I took a quick look at the repo, and that most probably should be transformed into a module that just provides the screenshot function.

See also #81.

PR #88, opened 2016-03-01
More render tests

henrikrudstrom:
depends on #84
adds tests for anchoring, 
AnchorsFill.qml fails, probably because it references a sibling.


ChALkeR:
Not sure why the existing tests png images were changed. Is that required. They were smaller (optimized), btw. Could you remove those changes, please?

The actual changes look good, but I would prefer to keep the opacity for now until we decide on that one.

Would it be fine to you if I would rebase this, remove changes present in #84 (those could be discussed separately) and add opacity for now?

tests/QMLEngine/scope.js
```diff
@@ -19,4 +19,10 @@ describe('QMLEngine.scope', function() {
     expect(parentItem.sum).toBe(6600);
     div.remove();
   });
+
+  it('can reference sibling items by id', function() {
+    var qml = loader('Parent', this.div);
```
ChALkeR: Does that file even exist?

henrikrudstrom: which file?

ChALkeR: SiblingParent.qml

ChALkeR: ScopeParent.qml

henrikrudstrom: should be loader(´Sibling´... good your eyes are working at least

henrikrudstrom:
strange, PropertyGrid test (the one with transparent reference image) still passes even when the screenshot code adds white background...

henrikrudstrom:
cleaned up and rebased, @ChALkeR, @pavelvasev, @Plaristote LGTY?

ChALkeR:
Rebased against master, optimized images, minor style fixes.
@henrikrudstrom Is everything ok?

ChALkeR:
Ok, LGTM. Landing.

PR #89, opened 2016-03-01
Refactor tests

henrikrudstrom:
moved the `div` creating and removal code to `beforeEach` and `afterEach`

just add `setupDivElement();` to your test

```
describe('QMLEngine.scope', function() {
  setupDivElement();
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "karma-coveralls": "^1.1.2",
     "karma-jasmine": "^0.3.7",
     "karma-phantomjs-launcher": "^1.0.0",
+    "karma-spec-reporter": "~0.0.24",
     "phantomjs-prebuilt": "^2.1.4"
   },
   "scripts": {
--- a/tests/QtQuick/Rectangle.js
+++ b/tests/QtQuick/Rectangle.js
@@ -1,23 +1,22 @@
 describe('QtQuick.Rectangle', function() {
+  setupDivElement();
   var loader = prefixedQmlLoader('QtQuick/qml/Rectangle');
+
   it('White', function() {
-    var div = loader('White');
+    var div = loader('White', this.div).dom;
     expect(div.innerHTML).toBe('');
     expect(div.style.backgroundColor).toBe('white');
     expect(div.offsetWidth).toBe(200);
     expect(div.offsetHeight).toBe(100);
     expect(div.clientWidth).toBe(200);
     expect(div.clientHeight).toBe(100);
-    div.remove();
   });
   it('Color', function() {
-    var div = loader('Color');
-    expect(div.style.backgroundColor).toBe('red');
-    div.remove();
+    var qml = loader('Color', this.div);
+    expect(this.div.style.backgroundColor).toBe('red');
   });
   it('Transparent', function() {
-    var div = loader('Transparent');
-    expect(div.style.backgroundColor).toBe('transparent');
-    div.remove();
+    var qml = loader('Transparent', this.div);
+    expect(this.div.style.backgroundColor).toBe('transparent');
   });
 });
--- a/tests/Render/runner.js
+++ b/tests/Render/runner.js
@@ -84,9 +84,10 @@
 
   Object.keys(tests).forEach(function(group) {
     describe('Render.' + group, function() {
+      setupDivElement();
       tests[group].forEach(function(test) {
         it(test.name, function(done) {
-          var div = loadQmlFile(test.qml);
+          var div = loadQmlFile(test.qml, this.div).dom;
           var result, expected, loaded = 0;
 
           var process = function() {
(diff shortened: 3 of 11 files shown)


ChALkeR:
Hm. This includes several unrelated commits. Could you please file several pull requests next time (for independant changes)? It would be easier to review that way.

tests/failingTests.js
```diff
@@ -13,6 +13,11 @@ window.failingTests = {
       'Javascript'
     ]
   },
+  QtQuick: {
+    Timer: [
+      'can set intervals'
```
ChALkeR: Is there such a test?

tests/common.js
```diff
@@ -22,6 +21,34 @@ function loadQml(src, opts) {
...
+  });
+}
+
+var customMatchers = {
+  toBeRoughly: function(util, customEqualityTesters) {
+    return {
+      compare: function(actual, expected, diff) {
+        var result = {
+          pass: actual > expected * (1 - diff) &&
+            actual < expected * (1 + diff)
```
ChALkeR: Style: should be 
```js
          pass: actual > expected * (1 - diff) &&
                actual < expected * (1 + diff)
```

If lint thinks that's invalid — inform me, please =).

tests/Render/runner.js
```diff
@@ -104,7 +105,7 @@
               fileName: test.group + '/' + test.name + '.png'
             });
             result.onload = process;
-            div.remove();
+            //div.remove();
```
ChALkeR: No need to leave commented out code.

tests/failingTests.js
henrikrudstrom: was renamed, will remove

henrikrudstrom:
image test is failing again

henrikrudstrom:
@ChALkeR split up the commit and rebased them all to master. if it LGTY i will merge

package.json
```diff
@@ -19,6 +19,7 @@
     "karma-coveralls": "^1.1.2",
     "karma-jasmine": "^0.3.7",
     "karma-phantomjs-launcher": "^1.0.0",
+    "karma-spec-reporter": "0.0.24",
```
ChALkeR: I don't think that it would break anything to use `~0.0.24` here instead.

tests/QtQuick/Rectangle.js
```diff
@@ -1,23 +1,23 @@
 describe('QtQuick.Rectangle', function() {
+  setupDivElement();
   var loader = prefixedQmlLoader('QtQuick/qml/Rectangle');
+
   it('White', function() {
-    var div = loader('White');
+    var qml = loader('White', this.div);
+    var div = this.div;
```
ChALkeR: Inconsistency between here and `Render/runner.js`. Also, `qml` in unused here.
Perhaps just `var div = loadQmlFile(test.qml, this.div).dom;`?

ChALkeR: Inconsistency between here and `Render/runner.js`. Also, `qml` in unused here.
Perhaps just `var div = loadQmlFile('White', this.div).dom;`?

tests/common.js
```diff
@@ -46,6 +54,9 @@ function loadQml(src, opts) {
                   ' is known to be failing. Skipping...');
       return;
     }
+    var div = document.createElement('div');
+    this.div = div;
```
ChALkeR: Style: `this.div = document.createElement('div');` instead of those two lines.

tests/common.js
```diff
@@ -46,6 +54,9 @@ function loadQml(src, opts) {
...
     itOrig.apply(this, arguments);
+    div.remove();
```
ChALkeR: Style: `this.div.remove()`.
This is more readable, because this way there is no alias for the variable and one doesn't have to keep in mind that `div` and `this.div` are indeed the same var.

ChALkeR: Aliasing is useful where the variable is used multiple times, but here it's just created and removed.

ChALkeR:
LGTM with style nits.

ChALkeR:
Re-opened due to #101.

ChALkeR:
I also technically don't see an LGTM from @Plaristote here, so I'm removing that one.

ChALkeR:
Ok, this fails after rebasing against master. Something is wrong here.

ChALkeR:
Btw, I updated this PR from 4d36ad1 and 979d61d.

ChALkeR:
`this.div` is undefined inside
```js
    expect(function() {
      var div = loader('NamedSignal', this.div).dom;
    }).not.toThrow();
```

ChALkeR:
That happens because `this` is different inside that context.

ChALkeR:
Update applied:
```diff
-    expect(function() {
-      var div = loader('NamedSignal', this.div).dom;
-    }).not.toThrow();
+    var div = loader('NamedSignal', this.div).dom;
```

PR #92, opened 2016-03-03
White background for test screenshots

henrikrudstrom:
the screenshot utility in qmlweb-dev has white background for all screenshots. the screenshot function in js sometimes creates transparent background. would like the outputs to match.
I dont believe there is any testcases we cannot test with white background, correct me if im wrong!
@ChALkeR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ build/
 *.kdev4
 *kate-swp
 .idea
+dev
+*.autosave
--- a/qmlweb-dev-install.bat
+++ b/qmlweb-dev-install.bat
@@ -1,0 +1,4 @@
+git clone https://github.com/qmlweb/qmlweb-dev.git dev
+cd dev
+build.bat
+cd ..
--- a/qmlweb-dev-install.sh
+++ b/qmlweb-dev-install.sh
@@ -1,0 +1,4 @@
+git clone https://github.com/qmlweb/qmlweb-dev.git dev
+cd dev
+./build.sh
+cd ..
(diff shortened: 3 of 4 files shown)


ChALkeR:
@henrikrudstrom Transparent background could be used for overlaying qml elements over something else, for example over html or WebGL scenes.
I'm not sure if we want to get rid of it, the tests are working as they are now.

Perhaps we _could_ make this check conditional (based on an env var), with opacity enabled by default and white bg enabled if an env var is present? This would allow to check with opaque screenshots at first, then generate images with transparency using `QMLWEB_SAVE_RENDER = 1` and use those.

What do you think?

Also, /cc @qmlweb/collaborators for opinions on this.

henrikrudstrom:
I see your case! My motivation was mostly pragmatic, not in the mood to dive into c++ atm. but i think it shouldnt be too hard to rewrite to transparent. 
For now perhaps the easiest solution is to not overwrite existing images and explicitly delete them when the test is changed.

ChALkeR:
Note that `qmlscene --transparent 1.qml` renders `1.qml` on _transparent_ background, and part of the window gets transparent.

You could see how that is done in http://code.qt.io/cgit/qt/qtdeclarative.git/tree/tools/qmlscene/main.cpp

ChALkeR:
@henrikrudstrom I don't think that we should compare only with white background, because that would create possible false positives for testcases where we would actually expect transparent background.

Given that both Qt QML and QmlWeb support transparent background, I suppose that we should keep that support and take that into an account when performing tests.

ChALkeR:
I am going to close this, as there were no activity for 1½ months after I stated that I would like to keep the opacity.

Feel free to reopen if you want to continue the discussion, though!

PR #94, opened 2016-03-03
Basic timer tests

henrikrudstrom:
depends #89

### DIFF
--- a/tests/QtQuick/qml/TimerSingleshot.qml
+++ b/tests/QtQuick/qml/TimerSingleshot.qml
@@ -1,0 +1,8 @@
+import QtQuick 2.5
+Timer {
+  interval: 50
+  property int counter: 0
+  onTriggered: {
+    yield("done")
+  }
+}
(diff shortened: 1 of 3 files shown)


Plaristote:
LGTM

tests/QtQuick/qml/TimerSingleshot.qml
```diff
@@ -0,0 +1,8 @@
+import QtQuick 2.5
+Timer{
```
ChALkeR: `Timer {`

ChALkeR:
LGTM

ChALkeR:
Re-opened due to #101.

ChALkeR:
Pushed changes from f44fe52, amended commit message.

ChALkeR:
`this.div` is undefined inside
```js
    expect(function() {
      var div = loader('NamedSignal', this.div).dom;
    }).not.toThrow();
```

PR #105, opened 2016-03-04
Item width or height set to zero.

igosoft:
Item with width or height set to zero can't rendered.


igosoft:
@ChALkeR  Please review and merge.

ChALkeR:
@igosoft This fails the tests for whatever reason.
Also, I am not sure about the changes at https://github.com/qmlweb/qmlweb/pull/105/files. Are those all of them? Perhaps you missed something?

igosoft:
@ChALkeR  reverted that one. However I don't understand what/where is failing but I'm sure that width/heigh default values are zero.

igosoft:
@ChALkeR  It probably fails due to NumberAnimation.qml test which as the root component that uses Item which doesn't have background. Am I right ?

PR #108, opened 2016-03-05
implicitHeight/implicitWidth take preference over width:0/height:0

igosoft:
For some unknown reason implicitWidth and implicitHeight doesn't behave as in Qt.


src/qtcore/qml/elements/Item.js
```diff
@@ -84,6 +84,7 @@ function QMLItem(meta) {
     this.heightChanged.connect(this, updateVGeometry);
     this.implicitWidthChanged.connect(this, updateHGeometry);
     this.implicitHeightChanged.connect(this, updateVGeometry);
+
```
ChALkeR: unrelated

igosoft: I'm working from web page :) (via github). That change is just to create a branch.

igosoft:
I'm working from web page :) (via github). That change is just to create a branch.

henrikrudstrom:
was wondering the same.. @igosoft have you checked out [cloud9](http://c9.io)

igosoft:
@henrikrudstrom looks great but I didn't work with yet

igosoft:
@ChALkeR  Please review and merge.

src/qtcore/qml/elements/Item.js
igosoft: @ChALkeR  Can I delete these comments. They can confuse now.

ChALkeR:
@igosoft Could you explain why `forceTheChange` is needed?

igosoft:
During construction phase we need to force to apply the change even if default value is the same as pointed by a user in qml file. In other words we must know listeners to know about set properties.

ChALkeR:
@igosoft Hm, makes sense.

Note that `src/qtcore/qml/elements/QtQuick/ListElement.js` also uses `applyProperties`. Should it use that parameter or not? (I am not sure).

ChALkeR:
Also note that `applyProperties` is recursive and atm does not propogate that flag somewhy. Is that intended? (I am not sure here either, just being precocious).

igosoft:
@ChALkeR I'm going to investigate your findings tonight ot tomorrow.

igosoft:
@ChALkeR thanks
1. ListElement - I confirm there is no need to "force" the change when ListElement is created (it doesn't have properties with values set by us in src code, so initial change is always propagated)
2. recursion - this can be fixed. It's needed for properties group and some of them are intil by us in sec code.

ChALkeR:
@igosoft Could you rebase this or file a new pull request with the same changes, please?
It's hard to understand anything here =).

ChALkeR:
Looks like the base commit for this is f77fd5e.

ChALkeR:
Actual new test extracted to https://github.com/qmlweb/qmlweb/tree/igosoft-implicit-width-height2. It fails atm.

ChALkeR:
The problem is that `width: 0` and `height: 0` do not override `implicitHeight`/`implicitWidth` as they should.

ChALkeR:
Testcase merged in bae45da.

Plaristote:
There has been fixed in ffbacb68159.
Properties having the `real` type are initialized to zero by default, meaning that declaring a height or width of zero didn't register as a change, thus not changing the size policy ($isUsingImplicitWidth/Height didn't get updated).
I overrode the initial value for an Item's height and width to `null` instead, using the `initialValue` option in the property's initializer.

PR #113, opened 2016-03-06
eliminate Undefined and replace references with undefined

ChALkeR:
This is actually a good change. You can't override `undefined` nowdays, so there is no use in saving it in a separate variable.


henrikrudstrom:
LGTM

Issue #107, opened 2016-03-05
Border width of Rectangle  can change the geometry of it.


igosoft:
Below example rendered in a browser changes size of Rectangle due to the fact that the size of the border width is greater than width of component.
  
Rectangle {
        border.width: 200   
        border.color: "red"
        height: 100
        width: 100
        color: 'blue'
 }

igosoft:
This one will be closed if #114 will be merged.

igosoft:
#114 resolved this one.

PR #114, opened 2016-03-06
rectangle with border.width=2 width=0

igosoft:
This test will check if no border is rendered for Rectangle when it doesn't have width height or implicitWidth implicitHeght and when its border.width is greater than zero.


igosoft:
Please review. Can I merge it ?

ChALkeR:
@igosoft Why was the second rectangle changed on the image?

igosoft:
Or why the 2nd rect had red border :)? In qml second rect doesn't have red border

ChALkeR:
@igosoft Damn. Sorry. It looks like my mistake when merging.

The border should be there, the border-radius shouldn't. `RectangleBorder.qml` should be a copy of `RectangleBorderChildren.qml`, but without the gray border on the parent element.

Not sure how did the tests pass.

Will fix ASAP.

ChALkeR:
Looks like `.getImageData()` returns only partial result — the top of the image is there, but everything from below point is just transparent black. Ow.

ChALkeR:
See #118 for the fix.
Btw, I guess this is blocked by #118.

ChALkeR:
@pavelvasev #118 landed, could you rebase this PR and redo the png, please?

igosoft:
@ChALkeR Please review and merge. I'm getting lost with my changes that start depend to each other :)

I had to add an extra if-condition checking Rectangle width and height when border with is set to zero. In html a border can exit without a content. I also merged #114 with #108 .

ChALkeR:
@igosoft Ah, it seems like you were merging between branches instead of rebasing, this is why the diff looks huge here.

I will rebase this.

ChALkeR:
@igosoft Rebased.

What was the «remove» meant to do?

src/qtcore/qml/elements/QtQuick/Rectangle.js
```diff
@@ -26,7 +26,11 @@ registerQmlType({
         this.css.borderWidth = this.border.width + 'px';
```
ChALkeR: Also, it's needed here.

src/qtcore/qml/elements/QtQuick/Rectangle.js
```diff
@@ -26,7 +26,11 @@ registerQmlType({
...
     });
     this.border.widthChanged.connect(this, function(newVal) {
-        this.css.borderWidth = newVal + 'px';
+        if (this.width > 0 && this.height > 0){
+            this.css.borderWidth = newVal + "px";
+        } else {
+            this.css.borderWidth = "0px";
+        }
```
ChALkeR: This won't work in any other case than zero-sized rectangle.

ChALkeR:
This change won't work with any other case than zero-sized rectangle.
Making the `border = Max(Max(width, height)/2, border)` would work when `Max(width, height)/2` is even, but wouldn't work when it's odd.

The suggested fix is:
```
QMLRectangle.prototype.$calculateBorder = function() {
  if (2 * this.border.width > this.width || 2 * this.border.width > this.height) {
    return '0 ' + '0 ' + this.height + 'px ' + this.width + 'px';
  }
  return this.border.width + 'px';
}
```

and use it to set the border from both those places.

igosoft:
@ChALkeR Looks good but what if rect.width or rect.heigth is equal zero ? and there is #107 - I have some ideas how to solve it.

ChALkeR:
This doesn't look ready yet.
Why was the testcase removed here?

ChALkeR:
@igosoft I rebased this. Could you please explain the «remove» commit (f68aca3)?

src/qtcore/qml/elements/QtQuick/Rectangle.js
```diff
@@ -12,31 +12,49 @@ registerQmlType({
     this.border = new QObject(this);
     createSimpleProperty("color", this.border, "color");
     createSimpleProperty("int", this.border, "width");
-
-    this.border.color = 'black';
+    this.border.color = "black";
```
ChALkeR: Unrelated.

src/qtcore/qml/elements/QtQuick/Rectangle.js
```diff
@@ -12,31 +12,49 @@ registerQmlType({
...
-    this.border.color = 'black';
+    this.border.color = "black"; 
     this.border.width = 1;
 
     this.colorChanged.connect(this, function(newVal) {
         this.css.backgroundColor = QMLColor(newVal);
     });
     this.radiusChanged.connect(this, function(newVal) {
-        this.css.borderRadius = newVal + 'px';
+        this.css.borderRadius = newVal + "px";
```
ChALkeR: Unrelated.

tests/Render/Simple/RectangleBorder.qml
```diff
@@ -1,37 +0,0 @@
...
-    border.width: 1
-  }
-  Rectangle {
-    x: 60
-    border.width: 0
-    border.color: 'red'
-    width: 10
-    height: 10
-  }
-}
```
ChALkeR: Why was the test removed?

src/qtcore/qml/elements/QtQuick/Rectangle.js
```diff
@@ -12,31 +12,49 @@ registerQmlType({
...
     });
 
     this.color = "white";
     this.radius = 0;
-    this.css.borderWidth ='0px';
-    this.css.borderStyle = 'solid';
-    this.css.borderColor = 'black';
+    this.css.borderWidth = "0px";
+    this.css.borderStyle = "solid";
+    this.css.boxSizing = "border-box";
```
ChALkeR: Not needed. All three lines here should be kept as they were.

src/qtcore/qml/elements/QtQuick/Rectangle.js
```diff
@@ -12,31 +12,49 @@ registerQmlType({
...
 
     this.$drawItem = function(c) {
+        //descr("draw rect", this, ["x", "y", "width", "height", "color"]);
+        //descr("draw rect.border", this.border, ["color", "width"]);
```
ChALkeR: What is this?

tests/Render/Simple/RectangleBorder.qml
igosoft: @ChALkeR I'm not sure but I think I removed not the file I wanted.

igosoft:
@ChALkeR Do you have any idea what is the reason of build failure ?

Plaristote:
@igosoft sounds like the parser couldn't parse qmlweb's source.
You can reproduce the issue using `npm run test`. The actual details of the parse error are unfortunately unavailable for now, it's one of the shortcomings of these JS_Parse_Error.
You can probably track it down using `git bisect`.

igosoft:
events.js:141
      throw er; // Unhandled 'error' event
      ^
Error
    at new **JS_Parse_Error** (eval at <anonymous> (/home/travis/build/qmlweb/qmlweb/node_modules/gulp-uglify/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:1526:18)

We have two **JS_Parse_Error** functions
1. src/uglify/parse.js 
2. src/qtcore/qml/lib/parser.js 

can that be a root cause ?

henrikrudstrom:
@igosoft made a pull request with a (perhaps temporary) fix to the Js_Parse_Error in #141

igosoft:
@henrikrudstrom  still the same :(

henrikrudstrom:
@igosoft This error came from the minification process (my fix was only for javascript in qml files), there was a syntax error in Rectangle.js fixed that, but some render tests still fail. 

PS if you run `QMLWEB_SAVE_RENDER=1 gulp test` the images rendered by phantom are saved in `tmp/`

igosoft:
@henrikrudstrom Thanks for the review/help/fix. I can fix this one tonight :)

ChALkeR:
Will take a look at this soon.

ChALkeR:
Squashed and cleaned up a bit.

ChALkeR:
This does need further fixes, but I'll merge this now, we can fix the remaining issues later.

ChALkeR:
Thanks!

PR #115, opened 2016-03-06
Merge pavel on destruction

henrikrudstrom:


src/qtcore/qml/QMLBaseObject.js
```diff
@@ -14,6 +14,10 @@ function QMLBaseObject(meta) {
     engine.completedSignals.push(this.Component.completed);
     this.completed = this.Component.completed;
 
+    this.Component.destruction = Signal([]);
+    //engine.destructionSignals.push(this.Component.destruction); would we need this?
```
ChALkeR: Could this be checked and sorted out somehow?

pavelvasev: @ChALkeR I think this is a case for graceful engine shutdown.

ChALkeR: @pavelvasev Are you proposing to pull this in as is (with the comment) and to sort it out later?
Would be fine by me, if the comment is changed to a TODO and a notice about graceful engine shutdown.

pavelvasev: @ChALkeR now I thinked again and it seems to be obvious, when we will decide to do graceful shutdown. So you might just remove that line for now.

src/qtcore/qml/QObject.js
```diff
@@ -26,5 +29,7 @@ function QObject(parent) {
 
         if (this.$parent && this.$parent.$tidyupList)
             this.$parent.$tidyupList.splice(this.$parent.$tidyupList.indexOf(this), 1);
+
+        this.parent = undefined; // must do this. => 1) parent will be notified and erase object from it's children. 2) DOM node will be removed.
```
ChALkeR: PTAL at change at https://github.com/qmlweb/qmlweb/commit/ebf0b623d298d2db8c95a0119ce4a980b5d882e1#diff-3a670166eb3912cd466ef8ce0543589d

Are those related? Should we do both or only one of them?

ChALkeR: /cc @pavelvasev

ChALkeR: /cc @igosoft

henrikrudstrom:
removed the commented line

src/qtcore/qml/QObject.js
henrikrudstrom: you mean `item.parent = undefined;`  [on line 19] (https://github.com/qmlweb/qmlweb/commit/ebf0b623d298d2db8c95a0119ce4a980b5d882e1#diff-3a670166eb3912cd466ef8ce0543589dR19)
that one is put at the end of the $delete() function on this branch (line33)

pavelvasev: I think here is a better place, because it guaranties that parent property will be cleared. In case of line 19, there is no guaranties because $delete might be called by someone else.

ChALkeR: @pavelvasev Does that mean that it would be unneeded there after this is merged?

ChALkeR: Looks so for me, but I would like you to confirm.

pavelvasev: @ChALkeR yes, I suggest to put `this.parent = undefined;` to the bottom of $delete method.

ChALkeR: Thanks!

ChALkeR:
LGTM

ChALkeR:
Thanks!

Issue #109, opened 2016-03-05
Timer tests periodically fail on Travis


ChALkeR:
/cc @henrikrudstrom 
Perhaps we should make them looser? E.g. <10% drift, not <5%.

PR #117, opened 2016-03-06
test undefined property

pavelvasev:


ChALkeR:
Travis failure is unrelated, I restarted it. Blame #109 for that.

ChALkeR:
@pavelvasev What's the actual result here? I mean — what does the property equal to?

pavelvasev:
@ChALkeR the idea is that if property value is not defined, it has the undefined value..

ChALkeR:
@pavelvasev I get that, I was asking what are the actual results atm.
Nevermind, I will check that.

ChALkeR:
Ah, it's `string` for some reason.

LGTM

PR #118, opened 2016-03-06
Fix render tests

ChALkeR:
There was an mistake where render tests were erroneously passing when they should fail.
That happened because `.getImageData()` was returning partial result (the image was not painted yet).

Also this fixes my merge error in the `RectangleBorder` render test.

/cc @henrikrudstrom @igosoft


ChALkeR:
Travis failure is unrelated, I restarted it. Blame #109 for that.

henrikrudstrom:
looks good

Issue #110, opened 2016-03-05
Add a css file?


ChALkeR:
I'm not sure if I like that idea or not, but let's have a discussion.
See #90 for the start of this.

The idea is to add a css file that could set the default styles to our elements based on classes.
This css file could also be split like our js files (per modules/components).

For example, we could apply `box-sizing: border-box` to all elements inside the root QML element this way.

Is it worth it? What would be the downsides of introducing a css file to the bundle?
/cc @qmlweb/collaborators.

igosoft:
Separating style from business logic is always good idea. However I'm not sure if we need css file for it. Shouldn't we control a style from Javascript ?

henrikrudstrom:
hmm, from a readablility perspective it would be nice if everything is defined in javascript, could we not use inheritance to set these properties?

ChALkeR:
> could we not use inheritance to set these properties?

Perhaps we could, looks like a good idea to me.

ChALkeR:
Approach like in #120 and setting everything from js based in iheritance instead of styles should work.
I'm closing this, if you have other opinions — feel free to reopen.

Issue #90, opened 2016-03-02
TextInput does not follow its height property


travellhyne:
The internal input element does not seem to follow the height of its container. It also overflows from the container.

 Setting the input to have a 100% height and `box-sizing` to `border-box` might be able to fix this.

**Without `{height: 100%; box-sizing: border-box;}`**
![image](https://cloud.githubusercontent.com/assets/1308095/13452163/805316f0-e080-11e5-8fd8-377a0a464387.png)

**With `{height: 100%; box-sizing: border-box;}`**
![image](https://cloud.githubusercontent.com/assets/1308095/13452191/af4ce8a0-e080-11e5-9777-dbc2e24d7198.png)

henrikrudstrom:
there seems to be other components that miss "box-sizing" see #72

Plaristote:
Considering the ways border work in QML, I assume `border-box` should be set for every element within a QML web page.
It's perhaps something that we shouldn't fix within TextInput or BorderImage, but set globally. Perhaps by using a stylesheet.

ChALkeR:
@Plaristote has a valid point here, perhaps we should add a «qml» class and assign it on all items (or just the root item). I do not think that we should change the _default_ behaviour, because one could mix qml and html.

But let's for now just set the styles manually as a quick fix, and discuss the addition of a css file separately.

ChALkeR:
@travellhyne Could you also provide a QML testcase, please?

ChALkeR:
Status update: c9a3d27 by @igosoft should fix the height (not yet merged), and we are discussing at #110 the best way to apply `box-sizing: border-box` to everything QML by default.

ChALkeR:
`box-sizing` part is at #120.

ChALkeR:
The `height` part is at #122.

ChALkeR:
@travellhyne, this should be fixed in [v0.0.7](https://github.com/qmlweb/qmlweb/releases/tag/v0.0.7).

I'm closing this now, feel free to ask for reopen if this wasn't fixed somewhy.

PR #120, opened 2016-03-06
Set box-sizing: border-box on all Items

ChALkeR:
This uses inheritance to ensure that all QML DOM items will have `box-sizing: border-box`.
It also modified `Initialize` tests to check that.

See #90 and #110.


Plaristote:
LGTM, let's merge this and finally close all of our border issues !

ChALkeR:
@Plaristote This will not close _all_ border issues, unfortunately: the placement of inner elements when border width is not zero would still need to be adjusted somehow.

PR #124, opened 2016-03-07
Uglybug fix of text input

henrikrudstrom:
fix updateloop for textinput
(659f38103b4b72dc3d74c5dd0183fc0301db4531)

### DIFF
--- a/src/qtcore/qml/elements/QtQuick/TextInput.js
+++ b/src/qtcore/qml/elements/QtQuick/TextInput.js
@@ -41,7 +41,11 @@ registerQmlType({
     });
 
     this.textChanged.connect(this, function(newVal) {
-        this.dom.firstChild.value = newVal;
+        // We have to check if value actually changes.
+        // If we do not have this check, then after user updates text input following occurs:
+        // user update gui text -> updateValue called -> textChanged called -> gui value updates again -> caret position moves to the right!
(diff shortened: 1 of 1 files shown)


Plaristote:
I'm fairly certain we've used this kind of behavior in other places.

This LGTM, but I wonder if we shouldn't fix that issue somewhere else (should we really trigger a propertyChanged event if the property has been set to the same value ?).

ChALkeR:
Qt QML does not fire on*Changed signals if the new value is equal to the old value.

ChALkeR:
I'm leaning to rejecting this PR and fixing this in the generic property logic instead.

ChALkeR:
Hm. Signals logic is fine. I'll check what's going on here.

ChALkeR:
Ah, I understand now why and how this happens.
@Plaristote It's not about propertyChanged:
 1. Create a TextInput, type some text into it.
 2. Move cursor to the middle of that text, type something.
 3. The cursor moves to the end of the text.

The signals are fired correctly, see #135, there is only one signal fired when the text is actually changed. But when it's changed, there is a listener that updates the DOM element value, and this one did not check if the current value is already equal to the one that is going to be set, and re-assigned it. It has nothing to do with our signals, it's DOM value updating. And a DOM value update on an input field moves the cursor to the end.

LGTM, merging.

PR #127, opened 2016-03-07
Repeater: remove unneeded line

ChALkeR:
Manual `.parent` unsetting after `.$delete()` is not needed anymore due to 66222bc6cfb361c07602b8581ca21940a9309d14.

/cc @pavelvasev


henrikrudstrom:
LGTM

PR #131, opened 2016-03-07
optimize png images

ChALkeR:
This is a lossless operation, actual content is not changed.


ChALkeR:
Saves 102 bytes on the logo, 400 bytes in tests.
Yes, they were already optimized =). I tuned the parameters a bit.

Plaristote:
LGTM

henrikrudstrom:
LGTM, @ChALkeR what tool are you using to compress, is that something we could put into the qmlweb-dev project?

ChALkeR:
@henrikrudstrom something like this:
```bash
#!/bin/bash
exiftool -overwrite_original -all= "$1";
pngcrush -blacken -brute -reduce -ow -l 9 "$1";
advdef -z -4 "$1";
```

I'm not yet sure if this is optimal, so I'm not certain if this should be the recommended way.

PR #137, opened 2016-03-09
travis: Report code coverage to codecov.io

### LINKED ISSUE #136: Use codecov.io instead of coveralls?
Coveralls was not very convenvient — it mixed up coverage results for the master branch and for PRs for some reason, it was hard to understand what was going on with the coverage score without looking into each individual PR manually.

On top of that, coveralls is down now.

I will give https://codecov.io/ a try, perhaps.

ChALkeR:
This is a work in progress, do not review yet.

Fixes  #136.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,7 @@ cache:
   directories:
     - $HOME/.npm/
     - node_modules
+before_install:
+  - npm install -g codecov
+after_success:
+  - codecov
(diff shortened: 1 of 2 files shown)


ChALkeR:
Looks to be blocked by splitting the coverage report by source files.

henrikrudstrom:
which depends on integrating browserify..
can we not use it with the compiled source as we did with coverall?

ChALkeR:
@henrikrudstrom It seems to refuse to even give out a score when the inspected file is not in the repo.

PR #139, opened 2016-03-09
test: cannot parse function assigned to var property

henrikrudstrom:
this works in qt qml but not in webqml: 
`property var aFunction: function(){}`


ChALkeR:
LGTM

ChALkeR:
This is now fixed in https://github.com/qmlweb/qmlweb-parser/commit/3fea94d53efb48f74062c90e61cf3f60496b7f96 and https://github.com/qmlweb/qmlweb/commit/448a3ded44dbc428c70cfe046b4d0158c143be82.

PR #141, opened 2016-03-09
print qml parse error in stacktrace

### LINKED ISSUE #140: code duplications: lib/parser.js ~ uglify/parse.js
The parser is currently a bit beyond me, there seems to be a *lot* of duplicate code or at least functions in `lib/parser.js` and `uglify/parse.js`. I suppose a lot of this can safely be removed

henrikrudstrom:
Prints the error like this:
```
1) QML TEST: SimpleRenderTest
     Test.TestCase
     Error: Unexpected token: punc (() (line: 6, col: 23, pos: 83)
  3      id: test
  4      delay: 200
  5
>>6      startTest: function(){
  7          test.compareRender("", function(equal){
  8              expect(equal).toBe(true);
  9              testDone();

 (line 7)
JS_Parse_Error@/home/henrik/Development/qmlweb/lib/qt.js:9:167246
js_error@/home/henrik/Development/qmlweb/lib/qt.js:9:167557
...
...
```

Both lib/parser.js and uglify/parse.js defines `Js_Parse_Error` so i had to delete one, not sure which one.. this will be temporary until we resolve #140

### DIFF
--- a/src/qtcore/qml/lib/parser.js
+++ b/src/qtcore/qml/lib/parser.js
@@ -269,25 +269,24 @@ function parse_js_number(num) {
         }
 };
 
-function JS_Parse_Error(message, line, col, pos, comment) {
-        this.message = message;
-        this.line = line + 1;
+function QMLParseError(message, line, col, pos, comment) {
+        line = line + 1
+        this.line = line;
         this.col = col;
         this.pos = pos;
         this.comment = comment ? comment : "";
+        this.message = message + " (line: " + line + ", col: " + col + ", pos: " + pos + ")" + "\n" + comment + "\n"
         try {
                 ({})();
         } catch(ex) {
                 this.stack = ex.stack;
         };
-};
 
-JS_Parse_Error.prototype.toString = function() {
-        return this.message + " (line: " + this.line + ", col: " + this.col + ", pos: " + this.pos + ")" + "\n" + this.comment + "\n" + this.stack;
 };
+QMLParseError.prototype = new Error();
 
 function js_error(message, line, col, pos, comment) {
-        throw new JS_Parse_Error(message, line, col, pos, comment);
+        throw new QMLParseError(message, line, col, pos, comment);
 };
 
 function is_token(token, type, val) {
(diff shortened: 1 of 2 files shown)


ChALkeR:
I don't think that we should modify `src/uglify/*` too much, it's hopefully going to be removed and replaced with a parser dependency on upstream uglify.

henrikrudstrom:
The other option is to rename Js_Parse_Error. I renamed it to QMLParseError. better name suggestion is welcome.

Plaristote:
This isn't from `src/uglify/*` though. There's two javascript parser in this lib right now, and the one that is being modified in this commit is the original one: while it uses parts of uglify, it has nothing to do with `src/uglify/*` (which *is* an unmodified snapshot of uglify, the kind of which we could get completely rid of and replace with a `require` call).

This however, is the parser from `qml/lib/parser.js`. Since there was changes made to it in order to support QML (the comment do say that the Javascript parts will change, though as little as possible), I'm not sure we can get rid of it so easily.

We might be able to split it, moving the parts that come from Uglify to a dependencies, and then use monkey-patching to get the same results. Considering the file was first introduced by Anton, I think we'll need his opinion on that (I call forth you, @akreuzkamp !).

Should we keep this, should we split it, do we remember why do we need it ?

Plaristote:
That looks like a pretty useful commit to me: now that the `uglify` folder is gone, can we agree on what to do with it ?

src/qtcore/qml/lib/parser.js
```diff
@@ -1463,7 +1462,7 @@ function qmlparse($TEXT, exigent_mode, embed_tokens) {
             var moduleName = S.token.value;
             var isDottedNotation = (S.token.type == "name");
             next();
-            
+
```
ChALkeR: Unrelated. The cleanup will be done later, after merges.

ChALkeR:
LGTM.

ChALkeR:
@Plaristote just to make sure — does this change look good to you?

Plaristote:
Yup, this LGTM.

ChALkeR:
Landed, thanks!

PR #142, opened 2016-03-09
Throw error instead of logging when Qml constructor is not found

henrikrudstrom:
Im not a big fan of failing silently, is there any reason not to throw more errors?

### DIFF
--- a/src/qtcore/qml/qml.js
+++ b/src/qtcore/qml/qml.js
@@ -155,8 +155,7 @@ function construct(meta) {
                 item.dom.className += " " + meta.object.$class + (meta.object.id ? " " + meta.object.id : "");
             var dProp; // Handle default properties
         } else {
-            console.log("No constructor found for " + meta.object.$class);
-            return;
+            throw new Error("No constructor found for " + meta.object.$class);
         }
     }
 


Plaristote:
I would say it might be useful in order to figure out whether a constructor is available or not. As long as we don't see an actual need for it, I see no reason not to merge this anyway. Though I think we should first get rid of line 159, it's unreachable now.

src/qtcore/qml/qml.js
```diff
@@ -155,7 +155,7 @@ function construct(meta) {
                 item.dom.className += " " + meta.object.$class + (meta.object.id ? " " + meta.object.id : "");
             var dProp; // Handle default properties
         } else {
-            console.log("No constructor found for " + meta.object.$class);
+            throw Error("No constructor found for " + meta.object.$class);
```
ChALkeR: `throw new Error`. Also, the `return` isn't needed.

Plaristote:
I took the liberty of fixing both issues (`throw new Error` and the unreachable `return`).
That looks good to me now.

ChALkeR:
LGTM

PR #146, opened 2016-03-11
replace js parsing with `eval()`  and remove uglify sources

henrikrudstrom:
Any reason we cannot just use `eval` to parse the javascript imported in QML? 
Is this tested enough? because it seems to work.... (see discussion in #140)
@Plaristote, @ChALkeR, @pavelvasev, @igosoft


Plaristote:
```
var imported = eval("function my_function() {}");
console.log(imported); // displays 'undefined'
```
I might be wrong, but from my understanding, this wouldn't work.

henrikrudstrom:
try wrapping it in paranthesis...
```
var imported = eval("(function my_function() {})")
```
The "ImportJavascript.qml" test passes....

ChALkeR:
@henrikrudstrom I am not sure if this would actually work, and I might need some more time to think about that. Also, we need to discuss various usecases, e.g. `.pragma library`.

But for now, could you please get rid of eval and use `new Function(…)` instead?

ChALkeR:
Btw, this will most likely not remove all BSD-2-Clause source from the project, because qml parser is based on the uglifyjs parser.

ChALkeR:
> The "ImportJavascript.qml" test passes....

It's blacklisted. And it doesn't pass.

henrikrudstrom:
> It's blacklisted. And it doesn't pass.
damn, i had a feeling something was too simple here!

Plaristote:
I've looked a bit into this (in the idea of getting rid of uglifyjs).
From what I've tested, it should be possible to use qml's javascript parser to achieve the same result.

```
qmlparse("function foo() {\
var lavar = 12; \
function ma_function() {\
  console.log('la function', lavar); \
}\
}")
```

The resulting tree contains all the data we need to import javascript files. With a little bit of work, we could directly access the qmlmethod walker, and we should be able to skip the part where I add the fake foo function. We'd just be generating the tree, collecting the top level symbols, and then doing whatever we're already doing with these.

So, is this better than using uglifyjs ? I'm not sure, but if it allows us to shrink our library's weight, I'm all for it.

henrikrudstrom:
Sounds like a worthwhile effort to me, getting rid of 2500+ lines of code. 
Also, we can then we are more flexible on how we handle qml-javascript import statements: `.import Qt.test 1.0 as JsQtTest`

akreuzkamp:
The reason why we need to parse js files is
```javascript
//test.js
var foo = "bar";
```
```javascript
import "test.js" as Test
Text {
    text: Test.foo
}
```
Must be supported and that can't be done with eval().
Don't know about uglify parser vs. qml parser, though.
@Plaristote What do you mean with "qml's javascript parser"? The qml parser is based on uglify as well or is there another one?

Plaristote:
There's uglify's parser used for javascript files, and then there's another parser based on uglify's one, that was used for qml files, if I'm not mistaken.
It shouldn't be too hard to get completely rid of uglify, and use the original parser for javascript files as well.

akreuzkamp:
You mean, rewriting all of the js-parsing work done by uglify?

Plaristote:
Yes, right now it's only 40 lines of code, asking uglify to retrieve the vars and function names in the top level scope (see `src/qtcore/qml/lib/jsparser.js`).
We could do the same thing without uglify, by walking through a tree generated by qml's js parser, instead of uglify's.

Plaristote:
I've done what I was talking about earlier in PR #153 (replace uglify's parse with qmlparse). If it gets merged, we'll have only one parser left to worry about.

PR #147, opened 2016-03-13
Qml Testcase element

henrikrudstrom:
Added a qml TestCase element that works in both qmlweb and qt. It defines a test function, handles startup delay and timeout, has basic test functions (`expect.toBe` and `fail` for now) and a `compareRender` function. 

When a test containing a `TestCase` element as the child of the root is run in jasmine:
1. a new spec is created
2. the test functions are mapped to the corresponding functions in jasmine
3. test function is run

[qmlweb-dev](https://github.com/qmlweb/qmlweb-dev) now has an utility `qtqmltest` which can load a qml file and run the testcases defined in it. The `compareRender` function creates a reference screenshot instead of comparing it. 

To test it: 
from the qmlweb directory: 

```
git clone https://github.com/qmlweb/qmlweb-dev.git dev
cd dev 
./build.sh
```

then run:  `node dev/render.js`


ChALkeR:
Sorry for the delay, will take a look at this now.

ChALkeR:
Hm. Perhaps we need a proper way to declare elements in qml first (with parsing them at the build-time).
TestCase element looks like something we would want to have in the library itself, not just in the tests folder.

ChALkeR:
This pull request is a little to big and covers different topics, not only the TestCase element. It would be a lot easier to review if this was split to independant changes.

henrikrudstrom:
@ChALkeR 
I agree, would be good to have a way to declare qml objects as part of the source. But we should make sure we can properly debug the code if we precompile the qml elements, or add the option to run them uncompiled for testing/debugging purposes. 
Perhaps we could start by allowing qml files in the source folder?

as for the commits, i know its a big one, but im not sure i have time to squash them again in the near future, but here is a description: 

82bac2de55c750da2656dd9f01d91fb48b113fbd: adds some file to gitignore (related to the screenshot generator)

aa70e8fc50c53c0bcf0cfa44f94152f2742ae8d6: `render.js`, basically the existing render code moved to a separate file and defines some global functions to be available in the qmltest: 
`window.compareScreenshot`, `window.renderTest`, `window.it`

83ac1beb12d1ee424b33e3c716a684885358f921: replace the existing render code with the refactored render code.

208172c3079d98067fdd8426c02282c08e8cc268: adds the TestCase qml element and some stylenits. Adds the option to specify additional import paths to the prefixedQmlLoader (not sure that is actually used in this PR), 

71b7088311163ba065669db6604d9e487b650cd4: adds the qml testcases, and an example testsuite, `tests.js`.

07a1bf7fee45ec2a804e6f71a0deb05a2081edc7: adds support for a `before()` function in a testcase to change properties before the qml element is rendered. and some minor tweaks.

Plaristote:
This has stalled long enough. Unless someone else objects, I'll take care of splitting that PR in the next days.

ChALkeR:
Related: #389.

ChALkeR:
I don't think we should make render tests more complex.
Instead, we should make a small testcase that could be used with Qt to check that all those tests pass on Qt, like `qmltestrunner -input tests/Auto/` for Auto tests.

ChALkeR:
> Instead, we should make a small testcase that could be used with Qt to check that all those tests pass on Qt, like `qmltestrunner -input tests/Auto/` for Auto tests.

That's done in cb3283c26a3bb05dde51ee26bbe32fd0f74029ea.

PR #148, opened 2016-03-21
using qrc with readQmlDir to cache qmldir files

Plaristote:
Allows gulp-qmlweb to aggregate the qmldir files within the qrc object.
No changes required in gulp-qmlweb for this to work, although I'll update the README to have the default file matchers look like this:
```
var qmlFiles = [ 'qml/**/*.qml', 'qml/**/*.js', 'qml/**/qmldir' ];
```
This is all that's needed to make this work with gulp-qmlweb.


ChALkeR:
LGTM

PR #149, opened 2016-03-21
adds array type to default constructor list

Plaristote:
Split from commit 201e20d. AFAIK, the only thing from that commit that should be merged.


henrikrudstrom:
LGTM

ChALkeR:
Is `array` type supported by QML?
I can't find it on http://doc.qt.io/qt-5/qtqml-typesystem-basictypes.html#basic-types-provided-by-the-qml-language

Plaristote:
Good point. I didn't find anything either. Turns out the examples from the same commit were using this type for things that didn't match the documentation.

I then wondered what type could be used, for instance with states. Variant is good enough for us, but I stumbled upon this:
http://doc.qt.io/qt-5/qml-list.html
and this:
http://comments.gmane.org/gmane.comp.lib.qt.qml/866

The parser currently can't work with the `property list<Item> slides`. Should we implement this ?

akreuzkamp:
list is supported and what you were looking for. Only defining own lists (`property list<Item>`) isn't, so far.

ChALkeR:
@akreuzkamp But `list` is already there =)

There is no reason to add an `array` type that isn't supported by Qt.

PR #152, opened 2016-03-28
Merge "properly remove animations at overwriting on init."

Plaristote:
> During initialization it can happen that a property gets assigned a
> binding and later, still during init, gets assigned a constant value.
> In this case it wouldn't properly remove the binding and when later,
> during evaluating bindings phase, the constant value would get
> overwritten again, by the binding. This is not intended.
> 
> Now it's cleaned correctly.

Apart from one change, this merge was just about moving some code around.

Though I found a single line of code that still used the property setter function while using the `fromAnimation` parameter instead of the `reason` parameter introduced in this commit. It's the change I made to, `src/qtcore/qml/elements/Item.js`, using QMLProperty.ReasonUser as a parameter. I figured it was the right value to use in that context.

### DIFF
--- a/src/qmlweb/RestModel.js
+++ b/src/qmlweb/RestModel.js
@@ -141,7 +141,7 @@ registerQmlType({
     function updatePropertiesFromResponseObject(responseObject) {
       for (var key in responseObject) {
         if (responseObject.hasOwnProperty(key) && self.$hasProperty(key)) {
-          self.$properties[key].set(responseObject[key]);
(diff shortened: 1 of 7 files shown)


ChALkeR:
`reason` looks to be required here (no default), but it is omitted in [src/qmlweb/RestModel.js#L144](https://github.com/qmlweb/qmlweb/blob/master/src/qmlweb/RestModel.js#L144).

ChALkeR:
RestModel has to be fixed. LGTM otherwise.

Plaristote:
Alright, I took care of `RestModel`.

PR #153, opened 2016-03-29
use qmlparse to parse javascript file, instead of uglify's parser

Plaristote:
This change should put an end to the confusion with `src/uglify/*` and `src/qtcore/qml/lib/parser.js`.
This commit gets rid of `src/uglify/*`, by using `qmlparse` instead of uglify's `parse` function to collect the symbols in a javascript file.

The `function foo` wrapper is necessary, as QML doesn't allow `var`s to be declared in the top level scope. So I create a function scope using the foo function, and then collect the symbols from that scope.


ChALkeR:
Hm.
Btw, how are we going to support `.pragma library`, `.import "something.js" as Something`, and `Qt.include('something.js`)?

It looks to me that this change would do no harm in that regard. Is that correct?

henrikrudstrom:
think its easier to implement those changes if we control the code of the parser. 
could it not be handled by text replacement and recursively calling `jsparse`:
for example, replace `.import "something.js" as Something` with `var Something = jsparse("something.js")` ?

henrikrudstrom:
LGTM

akreuzkamp:
We should not use qmlparse as it is, for that, imho.
I've made a commit to an extra branch makeParseGeneric. It makes the qmlparse method more generic and allows to parse js with it, so you don't need to put it inside a function and so we can add support for pragma, etc. later. Have a look at it and maybe rebase this PR on the commit. :)

Plaristote:
@ChALkeR: It doesn't affect it. @henrikrudstrom's idea sounds good to me as well, text replacement would be the easiest solution.

@akreuzkamp: I see, sounds like that would make this PR a lot more acceptable ! I'll check it out !

Plaristote:
@akreuzkamp I'm not sure you pushed the makeParseGeneric branch ! Or maybe you've done it to another repository, but I can't find it.

akreuzkamp:
Ah sorry, it's called makeParserGeneric (forgot the r of parser): https://github.com/qmlweb/qmlweb/tree/makeParserGeneric

Plaristote:
Alright, I've rebase this on the makeParserGeneric branch, and indeed, now it works without the function wrapper. I think we've reached something that's acceptable here.

I took the liberty of fixing a typo in the code (QMLDocument written as QmlDocument at one point, making qmlweb_parse.QMLDocument evaluate to undefined). Otherwise, the commit making QML Parser more generic LGTM.

ChALkeR:
LGTM.

We can always revert this later, if there would be some reasons to do that (I don't see any atm).

ChALkeR:
Nice, thanks!

PR #155, opened 2016-04-04
gulpfile: Clear QT_QPA_PLATFORM env var for PhantomJS to work

ChALkeR:
`QT_QPA_PLATFORM=xcb` is set by default in some systems now, e.g. Arch Linux, and that causes PhantomJS to fail.

Ref: https://github.com/ariya/phantomjs/issues/14061

This could be reverted after it gets fixed on PhantomJS side.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -29,6 +29,10 @@ var tests = [
   'tests/**/*.js'
 ];
 
+// This is required because other values confuse PhantomJS, and are sometimes
+// set by default by the system.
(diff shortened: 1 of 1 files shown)


ChALkeR:
This change is trivial, not related to the sources, and it looks like the problem is not affecting anyone else here but me.

I'm going to merge it if there would be no objections in the next few days.

Plaristote:
No objections and LGTM. I haven't experienced the issue but I can reproduce it by settings the env var myself.

PR #156, opened 2016-04-04
gulpfile: Cleanup qtcoreSources

ChALkeR:
This removes the `src/uglify/*` leftovers from the gulpfile (ref: #153) and merges `src/qtcore/*.js` into a single entry, they are all independant.

Note that `src/qtcore/*.js` is not the same as `src/qtcore/**/*.js` and is limited to a single level.


Plaristote:
LGTM

Issue #49, opened 2016-02-19
Merge KDE review 117456 — MouseArea.pressedButtons


ChALkeR:
See https://git.reviewboard.kde.org/r/117456/

Author: shry15harsh <shry.harsh@gmail.com>

Title:
> Add property 'pressedButtons' to MouseArea

Description:
> Added pressedButtons property for QMLMouseArea for QMLWeb. Currently, e.buttons is not available in browsers other than Firefox Dom Level 3. So, it is just assigned to mouse.button for now. So, this property will not be able to tell all the mouse buttons pressed at a time.

Diff: https://git.reviewboard.kde.org/r/117456/diff/raw/

ChALkeR:
PR: #157

ChALkeR:
PR merged.

PR #157, opened 2016-04-04
Merge «Add property 'pressedButtons' to MouseArea»

ChALkeR:
See #49.

/cc @shry15harsh 

This doesn't support multiple buttons yet, but LGTM.

Multiple buttons are supported by Firefox, Chrome (since 43) and Edge nowdays, so this change could be extended to determine if `.buttons` is present and fallback to the implementation currently proposed by this PR.


ChALkeR:
This doesn't support multiple buttons yet, but LGTM.

Multiple buttons are supported by Firefox, Chrome (since 43) and Edge nowdays, so this could be continued to determine if `.buttons` is present and fallback to the implementation currently proposed by this PR.

ChALkeR:
LGTM. This change is additive and looks valid, so I am going to merge soon if there will be no objections.

Plaristote:
No objections here, this looks fine to me.

Issue #143, opened 2016-03-10
Clean up folder structure


henrikrudstrom:
The folder structure in `src/` seems a bit random. Would be nice to clean this up to anticipate splitting the repo into modules, #2 and make the structure a bit flatter.

1. `src/qmlweb/` --- is this used, should it be part of the core project? 

2. `src/helpers/` and `uglify`--- will probably be removed soon anyway.

3. `qml.js` / `UpdateGeometry.js` / `QML*.js` --- these map nicely to the [QDeclarative](http://doc.qt.io/qt-4.8/qtdeclarative-module.html)

4. `Component.js`/`Item.js`/`PropertyChanges.js` --- dont they belong in the QtQuick folder? or they could go together with the `QML*.js` as they are part of the 

5. `qtcore/qml/lib/*.js` --- this is mostly parsing. `import.js` should be considered in #126.

6. `qtcore/*.js` --- mostly constants, not so much code. not sure if this should have a dedicated folder or be grouped together with qml/qdeclarative (3.)

7. `qtcore/qml/elements/*` --- could be moved a bit up the hierarchy

Just an initial draft, 
```
src/
   parse/ --- (5.)
   qtcore/ --- (6.)
   declarative/ --- (3. / 4.) not sure about the name (could be qml or qdeclarative or engine...)
   elements/ --- (7.) could also be called `modules`
      QtQuick/
      QtQuick.Controls/
      ...
  
   
   


```

Plaristote:
1. No, shouldn't be part of the core project.
However, the Settings object is supposed to be an implementation of [this](http://doc.qt.io/qt-5/qml-qt-labs-settings-settings.html), but back in those days, property aliases didn't work as they should, so I didn't put it in the right module. 
So maybe we want this object to go somehwere else. The RestModel isn't something from QML and should be removed.

2. Fairly certain I got rid of them in the `browserified` branch, so yeah, it's gonna disappear.

6. I think it needs its own folder. Ideally, we would have a part of qmlweb that could be used without qml altogether (all the features re-implementing things from Qt, rather than QML), and that's where these things would be.
This includes things such as the Qt object and the list models.

7. Yes, although I think they should all go in separate modules. Even the most basic QtQuick imports should probably be a separate module.

henrikrudstrom:
1. I see, persistent settings saved in xml or windows registry. would cookies be a good webqml analog?

3. "be used without qml altogether", Do you mean without any qml parsing related code?

[...]

PR #160, opened 2016-04-05
Add support for attached properties.

ChALkeR:
*Update: sorry, the descrption was from another PR at first.*

The source for this is 6022424e09f7692a664c1a31fa251ad2e0bd8514.

/cc @akreuzkamp, @Plaristote, @pavelvasev

### DIFF
--- a/src/qtcore/qml/elements/Component.js
+++ b/src/qtcore/qml/elements/Component.js
@@ -7,6 +7,17 @@ function QMLContext() {
     }
 }
 
+QMLComponent.getAttachedObject = function() { // static
+    if (!this.$Component) {
+        this.$Component = new QObject(this);
+        this.$Component.completed = Signal([]);
+        engine.completedSignals.push(this.$Component.completed);
+
+        this.$Component.destruction = Signal([]);
+    }
+    return this.$Component;
+}
+
 QMLComponent.prototype.createObject = function(parent, properties) {
     var oldState = engine.operationState;
     engine.operationState = QMLOperationState.Init;
--- a/src/qtcore/qml/QMLBaseObject.js
+++ b/src/qtcore/qml/QMLBaseObject.js
@@ -8,13 +8,6 @@ function QMLBaseObject(meta) {
     this.$isComponentRoot = meta.isComponentRoot;
     this.$context = meta.context;
 
-    // Component.onCompleted
-    this.Component = new QObject(this);
-    this.Component.completed = Signal([]);
-    engine.completedSignals.push(this.Component.completed);
-
-    this.Component.destruction = Signal([]);
-
     // Component get own properties
     var attributes = [];
     for (var key in meta.object) {
--- a/src/qtcore/qml/QMLEngine.js
+++ b/src/qtcore/qml/QMLEngine.js
@@ -632,5 +632,11 @@ QMLEngine = function (element, options) {
             options.debugConsole.apply(undefined, args);
         };
     }
+
+    // TODO: Move to module initialization
+    for (i in constructors) {
+        if (constructors[i].getAttachedObject)
+            setupGetter(QMLBaseObject.prototype, i, constructors[i].getAttachedObject);
+    }
 }
 


ChALkeR:
Perhaps `this.destruction = this.Component.destruction;` and `this.completed = this.Component.completed;` are not needed here? Those weren't present in the original commit base but were present in master, so I added them. Looking at this again, I think that those are redundant.

Can anyone confirm that?

src/qtcore/qml/elements/Component.js
```diff
@@ -7,6 +7,20 @@ function QMLContext() {
     }
 }
 
+QMLComponent.getAttachedObject = function() { // static
+    if (!this.Component) {
```
akreuzkamp: We need `this.$Component` here. That's because `this.Component` is a getter-setter-property and accessing that would result in this very function to be called again, causing an infinite loop ending in a stack-overflow.

akreuzkamp:
I don't understand at all, why we have this.completed and this.destruction. They are non-standard and not needed.

src/qtcore/qml/elements/Component.js
ChALkeR: Does that mean that `this.$Component` and `this.Component` are needed?

ChALkeR:
@akreuzkamp `completed` is since fff8a61ae93a7eb25480001edaf42b745db68845, `destruction` is since 5cd2a2297d61c7bb8422157a3331cb73487485e7 (looks like that one has a mistake, mixing `$Completed` and `Completed`).

I think it would be safe to remove those in a separate PR, and then rebase this on top of that.

ChALkeR:
@akreuzkamp Also take a look at 7d0ace1e — it adds `this.completed` to yet another place.

src/qtcore/qml/elements/Component.js
akreuzkamp: yes, `this.$Component` is needed to store the actual attached object, `this.Component` is the required public API and is just a "link" to `Component.getAttachedObject()`. I'm fine though with any other way to store the actual attached object. It's completely internal and I'm fine if it's only accessible via getAttachedObject (then we need some hasAttachedObject or get-but-don't-create-if-non-existant kind of function)

ChALkeR:
@akreuzkamp Updated to `$Component`, like it was in your commit. The only difference is the addition of `this.$Component.destruction = Signal([]);` (and the removal of `this.completed` / `this.destruction`).

ChALkeR:
It looks like the tests now fail at [tests/Initialize/runner.js#L76](https://github.com/qmlweb/qmlweb/blob/ca55b5ae58069b0339d524e682083375a365a56a/tests/Initialize/runner.js#L76)

akreuzkamp:
mhh, I guess it fails due to the module management. the original commit is against kde-master, which doesn't know modules. The part in QMLEngine.js probably needs adjustment. @Plaristote Can you maybe have a look at it? I don't know the module management code well enough, yet.

Plaristote:
As far as I can tell, it doesn't seem to come from the module management. I've inspected it, and the first item to fail, Column, actually gets constructed properly.

I have no idea why we can't access it through `qml.Component` yet, but there's no doubt that the component has been constructed.

Note that all the components failing are the specializations of QMLPositioner. It looks like it's related to the issue somehow. Only the QMLPositioners, and all of them are failing in this test.

ChALkeR:
This is caused by execution order of those components registration.

`inherit(QMLPositioner, QMLItem);` from QMLPositioner.js is called _after_ `registerQmlType` of everything that depends on QMLPositioner.

ChALkeR:
We could patch this by fixing the order in gulpfile. I would prefer a better solution, though, e.g. delaying `registerQmlType` calls.

ChALkeR:
@akreuzkamp @Plaristote Failures are temporary fixed by hard-coding the order in `gulpfile.js`.

I think that this would be better solved in the future, with #143 and dependency checking upon module loads.

The original PR LGTM.

ChALkeR:
@akreuzkamp On another hand, could you please provide an explanation in which usecase is this better than just storing `Component` directly?

akreuzkamp:
It's necessary, because any type can add attached properties to any object.Think of QtQuick-Control's Stack type adding an attached object to every Item inside a StackView. We can't hardcode that.

Another advantage is memory consumption. All Items have an attached property called Keys, providing lots of properties and signals for keyboard management, which aren't used for most of them. This commit actually attaches attached-objects (on-demand) and thus saves some memory. :)

Plaristote:
Sounds good to me. Should we merge this ?

ChALkeR:
Yes, LGTM.

ChALkeR:
It looks like an older version of this was merged. I updated this PR some time ago.

This misses d29c2c92c11c50ed0fdfd263a0e07b42ad36d956, which fixed the issues.

Plaristote:
Alright, I see what happened here, I'm going to close that pull request again in a few seconds.

PR #161, opened 2016-04-05
Repeater: pass on the Component.completed() signal to all children.

ChALkeR:
Extracted from 2865c68cd7b5fc4865b8db474c1d9cc75eaf8c23.

Comment by @akreuzkamp:
> $tidyupList is a list of internal children (compare to QObject::children() in contrast to QQuickItem::children()!). It's called tidyupList because it's main use is to remove references to child objects when deleting an object. The whole concept of memory management QmlWeb uses so far is in great need of improvement.
In short: It passes on the Component.completed() signal to all it's children.

I still don't understand this completely, and this needs a review by someone.

/cc @akreuzkamp, @Plaristote, @pavelvasev

### DIFF
--- a/src/qtcore/qml/elements/QtQuick/Repeater.js
+++ b/src/qtcore/qml/elements/QtQuick/Repeater.js
@@ -25,8 +25,9 @@ function QMLRepeater(meta) {
 
     function callOnCompleted(child) {
         child.Component.completed();
-        for (var i = 0; i < child.children.length; i++)
-            callOnCompleted(child.children[i]);
+        for (var i = 0; i < child.$tidyupList.length; i++)
+            if (child.$tidyupList[i] instanceof QMLBaseObject)
+                callOnCompleted(child.$tidyupList[i]);
     }
     function insertChildren(startIndex, endIndex) {
         for (var index = startIndex; index < endIndex; index++) {


Plaristote:
I would review this, but I'm not sure about ·$tidyupList either. I don't see how it differs from $children.

akreuzkamp:
@Plaristote It's simple: $tidyupList exists, $children does not :P
Really: $tidyupLists contains some more stuff than just children, but for this purpose it's just like list of children. I'm just preparing a commit to split up tidyupList into children, signalConnections and stuff.

akreuzkamp:
lgtm, currently it's going to fail, because the inheritance doesn't work correctly. But that's an independent problem, that I'm hunting right now, so I think this commit can be shipped. :)

ChALkeR:
@akreuzkamp A quick question to ensure that I got everthing correct — are all members of `.children` also present in `.$tidyupList`?

akreuzkamp:
Not necessarily. The only `.children` property we have is Item's children property, which is about *graphical* children. The children I meant with `.$children` is logical children.
I.e. if you have
```
Row {
    Repeater {
        delegate: Item { id: foo }
    }
}
```
foo will be a graphical child of the Row but a logical child of the Repeater. Although Repeater is an Item (which doesn't make any sense to me) its children property will always be empty.

.$tidyupList contains logical children. But in the end it should come to the same result. :)

ChALkeR:
@akreuzkamp Ah, I get it now, thanks.

LGTM, landing.

PR #166, opened 2016-04-08
Better error show

Gaubee:
Assigned to the error column

```
Unexpected token name undefined, expected punc { (line: 2, col: 28, pos: 47)
  1  import QtQuick 2.6
>>2  import "E:/Git/中文测试/测试啊" __ as HTMLBase
                    　　　　 　　　     ∧
  3  HTML{
  4     Head{
  5         id: qaq

TypeError: (intermediate value) is not a function
```

Signed-off-by: Gaubee gaubeebangeel@gmail.com

### DIFF
--- a/src/qtcore/qml/lib/parser.js
+++ b/src/qtcore/qml/lib/parser.js
@@ -303,12 +303,12 @@ var cjk = function(char_code) {
         0xF900 <= char_code && char_code <= 0xFAFF ||
         0x2F800 <= char_code && char_code <= 0x2FA1F;
 };
-function replacerConsoleRange(str) {
+function fillWithWhitespace(str) {
     var res = "";
     for (var i = 0, len = str.length, cc; i < len; i += 1) {
         cc = str.charCodeAt(i);
         if (cc === 9) { // \t
-            res += '\t' // tab不管，否则不同的控制台下tab的长度是不一样的；除非在源码那边也进行修改
+            res += '\t' // Ignore TAB, or else in a different console TAB length is not the same. ~~Unless there are modifications in the source code, the source TAB replaced by a fixed length spaces~~
         } else if (cjk(cc)) {
             res += cjk_c
         } else {
@@ -317,32 +317,10 @@ function replacerConsoleRange(str) {
     }
     return res;
 };
-function repeatString(_str, count) {
+// reference from [MDN | String/repeat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat).
+// Internal use only, ignoring the type of judgment, increase speed
+function repeatString(str, count) {
     'use strict';
-    if (_str == null) {
-        throw new TypeError('can\'t convert ' + _str + ' to object');
-    }
-    var str = '' + _str;
-    count = +count;
-    if (count != count) {
-        count = 0;
-    }
-    if (count < 0) {
-        throw new RangeError('repeat count must be non-negative');
-    }
-    if (count == Infinity) {
-        throw new RangeError('repeat count must be less than infinity');
-    }
-    count = Math.floor(count);
-    if (str.length == 0 || count == 0) {
-        return '';
-    }
-    // Ensuring count is a 31-bit integer allows us to heavily optimize the
-    // main part. But anyway, most current (August 2014) browsers can't handle
-    // strings 1 << 28 chars or longer, so:
-    if (str.length * count >= 1 << 28) {
-        throw new RangeError('repeat count must not overflow maximum string size');
-    }
     var rpt = '';
     for (;;) {
         if ((count & 1) == 1) {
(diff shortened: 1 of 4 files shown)


Plaristote:
Correct me if I'm wrong, but check out PR #141. Is this trying to implement something similar ?

Gaubee:
@Plaristote Not the same, # 141 in improving the error stack, and this marked the column where the error.
If there is a complex expression. This SHOW will help developers to better position their mistakes

src/qtcore/qml/lib/parser.js
```diff
@@ -293,20 +293,104 @@ function js_error(message, line, col, pos, comment) {
...
+        count >>>= 1;
+        if (count == 0) {
+            break;
+        }
+        str += str;
+    }
+    // Could we try:
+    // return Array(count + 1).join(_str);
+    return rpt;
+};
```
akreuzkamp: To be sure: Is all of the code your copyright or did you copy this snippet from the web? In case of the latter, that's no big deal, but we need to know the license :)

src/qtcore/qml/lib/parser.js
```diff
@@ -293,20 +293,104 @@ function js_error(message, line, col, pos, comment) {
...
+var normal_c = " ";
+var cjk_c = "　";
+var cjk = function(char_code) {
+    return 0x4E00 <= char_code && char_code <= 0x9FFF ||
+        0x3400 <= char_code && char_code <= 0x4DFF ||
+        0x20000 <= char_code && char_code <= 0x2A6DF ||
+        0xF900 <= char_code && char_code <= 0xFAFF ||
+        0x2F800 <= char_code && char_code <= 0x2FA1F;
+};
+function replacerConsoleRange(str) {
```
akreuzkamp: I guess this is supposed to be named `replaceConsoleRange` right? :)

src/qtcore/qml/lib/parser.js
```diff
@@ -293,20 +293,104 @@ function js_error(message, line, col, pos, comment) {
...
+    var res = "";
+    for (var i = 0, len = str.length, cc; i < len; i += 1) {
+        cc = str.charCodeAt(i);
+        if (cc === 9) { // \t
+            res += '\t' // tab不管，否则不同的控制台下tab的长度是不一样的；除非在源码那边也进行修改
```
akreuzkamp: What's the meaning of this comment? :D

src/qtcore/qml/lib/parser.js
```diff
@@ -293,20 +293,104 @@ function js_error(message, line, col, pos, comment) {
...
-      var mark = ( i == line ) ? ">>" : "  ";
-      r += mark + i + "  " + lines[i] + "\n";
-  }
-
-  return r;
-}
+// CJK characters in the console display longer lengths, calculate the length of a string of actual display
+var normal_c = " ";
+var cjk_c = "　";
+var cjk = function(char_code) {
```
akreuzkamp: Maybe call this function `isCjk`? :)

src/qtcore/qml/lib/parser.js
```diff
@@ -293,20 +293,104 @@ function js_error(message, line, col, pos, comment) {
...
+    var codeLines = code_text.split("\n");
+
+    // Error line number index
+    var show_line_index = line;
+    // The starting line of the display code number
+    var show_start_line = Math.max(show_line_index - line_range, 0);
+    // End line of code displays the number of
+    var show_end_line = show_line_index + line_range;
+
+    // The number of characters required to display line, blank spaces + 1
```
akreuzkamp: it took me some time to understand, that you mean "line number" here. Maybe you can add it for clarity. :)

akreuzkamp:
Hey Gaubee,
thanks a lot for your contribution. It looks good to me. :)
I only have some minor nitpicks, though (only function naming and comments). Please have a small look at them. :)

src/qtcore/qml/lib/parser.js
Gaubee: copy from [MDN | String/repeat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat).

akreuzkamp:
looks good to me :)
@ChALkeR is the mdn code ok? Because of licensing.

ChALkeR:
From https://developer.mozilla.org/en/docs/MDN/About#Copyrights_and_licenses:

> Code samples added to this wiki before August 20, 2010 are available under the MIT license; you should insert the following attribution information into the MIT template: "© <date of last wiki page revision> <name of person who put it in the wiki>".

> Code samples added on or after August 20, 2010 are in the public domain. No licensing notice is necessary, but if you need one, you can use: "Any copyright is dedicated to the Public Domain. http://creativecommons.org/publicdomain/zero/1.0/".

ChALkeR:
That code at MDN was added [in 2014](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat$compare?to=648865&from=647035), which means that it's in the public domain, no license notices needed.

ChALkeR:
I abstain from reviewing the actual changes here, I'm not familiar with our parser =).

src/qtcore/qml/lib/parser.js
```diff
@@ -293,20 +293,92 @@ function js_error(message, line, col, pos, comment) {
...
+            rpt += str;
+        }
+        count >>>= 1;
+        if (count == 0) {
+            break;
+        }
+        str += str;
+    }
+    return rpt;
+};
```
ChALkeR: Btw, that code at MDN isn't actually good. It has an infinite loop and shift, is barely readable for the sake of optimizing performance, which I think it also fails to do.

In js, strings are immutable and pooled. Adding something to a string creates a new string in the pool, which is btw dependant on the previous string (because it consists of two parts — the old one plus the newly added one).

Constructing a string with concatenations of a single character is a good way to consume lots of memory in js =).

Though that's most likely not very important here (we don't expect lots of huge strings), I would just use `Array(count).fill(str).join('')` for the sake of readability.

ChALkeR: See https://github.com/pieroxy/lz-string/issues/46, for an example.

ChALkeR: Ah, above all, `for (;;) {` could be a deopt.

ChALkeR:
Note that this should now retarteted to https://github.com/qmlweb/qmlweb-parser, parser was moved there.

Plaristote:
I've taken a look at this to see if it was possible to move the PR to the new repository.
Really, it sounds like we would have to rewrite pretty much everything (completely everything if we're not keeping the MDN code).

Should we close this PR and open an issue on the new repository ?

ChALkeR:
@Plaristote Not sure. About the MDN code — it's ok to take it (it's public domain), but I see no need in that, as `Array(count).fill(str).join('')` would do pretty much the same thing and I doubt there would be any measurable performance loss, especially given that it's an error reporting function =).

ChALkeR:
No activity here, closing.
Moved as an issue to https://github.com/qmlweb/qmlweb-parser/issues/28/.

Issue #163, opened 2016-04-08
Merge fork-arnopaehler, take 2


ChALkeR:
This consists of everything that is left in #17 minus the unwanted files and style-only changes.

First set of patches (92eb17d split): [fork-arnopaehler-patches1](https://github.com/qmlweb/qmlweb/commits/fork-arnopaehler-patches1). Commits:
- [x] 5252956 Implement logCss() function — **Rejected**, unused debug method.
- [x] a712120 Implement TextEdit — **Merged** in #174.
- [x] 5feb0bb Implement TestPlugin — **Rejected**, this is non-standard, looks like a playground. We don't need this.
- [x] 79c6678 .gitignore: Add .idea/ and /.project — **Merged** in #174.

Second set of patches (6830c351056f16cc4f9a57060e7a1ced7964d846 split):  [fork-arnopaehler-patches2](https://github.com/qmlweb/qmlweb/commits/fork-arnopaehler-patches2). Commits:
- [x] 2585cb5 Row: register effectiveLayoutDirection — **Rejected**, it is never set here (this is a stub), and this is not the place where should be registered — it should be present for other positioners too, e.g. ListView, Flow, Grid, etc. 
- [x] fb6cd39 Add Qt.Align* constants — **Merged** in #167.
- [x] e7e6523 Implement Screen — **Merged** in #167.
- [x] 384b250 parser: walker updates — **Rejected**, see https://github.com/qmlweb/qmlweb/issues/163#issuecomment-208813734. 
- [x] 0a09a8c More work on Loader — **Merged** in #167.
- [x] 41f545e A clarification of a TODO comment — **Rejected**, inapplicable. That hack for older browsers was removed.
- [x] ba8b9ff Register Array as 'array' basic QML type — **Rejected**, see #149.
- [x] 8fe2ad9 Implement TextField — **Merged** in #174.
- [x] 79e6124 TextArea: use strict comarison in updateValue — **Rejected**, inapplicable, and I don't see why we would want this. In @arnopaehler fork, TextEdit doesn't have this change.
- [x] 1bded10 Implement objItems() and typeName() — **Rejected**, unused debug methods.
- [x] e7cee0d Default to 'Object List' in objList() title argument — **Rejected**, unused debug methods.
- [x] 36f6b2b SequentialAnimation: remove debug, remove descr() debug function — **Already done**, in #29 (63f0a081a62136239305faf07e115d7193e863a3).
[...]

PR #167, opened 2016-04-09
Merge some of arnopaehler patches

ChALkeR:
For the new elements implemented by @arnopaehler (Loader, Screen, ComboBox) this should be considered as an initial implementation, we can fix all the issues later, so lets not hold the merges =). I fixed a few trivial issues, though (in separate commits).

The Screen implementation is not valid here (hence my TODO comment), but could serve as a good starting point.

Ref: #163


Plaristote:
LGTM. Should we add the Screen and Canvas elements to the #1 issue ? Just to make sure we think of finishing them before 1.0.

ChALkeR:
@Plaristote Added.

Issue #32, opened 2016-02-17
Merge fork-pavelvasev


ChALkeR:
Merge fork by @pavelvasev.
That is actually based on #19, I excluded the commits that are already present in #19.

- [x] c567f68 README.md: Fix color name in tutorial — **Rejected**, Readme-only change, inapplicable.
- [x] **[listmodel]** 04bd0b1 ListModel: Notify dependent properties about list element changes with .set(). — **Merged** in #206.
- [x] **[upflow]** 40895fb upflow option — **Discarded** by @pavelvasev.
- [x] **[upflow]** 88839ab did it! up flow contexts in qmlweb! — **Discarded** by @pavelvasev.
- [x] **[context]** a623154 context comments — **Discarded** by @pavelvasev. Comments only.
- [x] **[aliaschangeback, textarea, textinput, button]** dce7d96 fixes from viewlang — Button changes are **discarded** by @pavelvasev, TextInput  signal changes **rejected** (not compatible with Qt QML). TextEdit padding fixes — **Merged** in #208. Alias fixes — **Merged** in #195.
- [x] **[checkbox]** 6bbfa83 checkbox fixes — **Merged** in #198.
- [x] **[video]** 816f334 fix crop mode for video — **Rejected**, can't reproduce any issue without it, @pavelvasev doesn't remember what that fixes. Perhaps that is controls-related, which we don't need.
- [x] **[recursiveprops]** 595a29f recursive mode for $initializePropertyBindings — **Merged** in #233.
- [x] **[createqmlobject]** a5efa1f createQmlObject fix — **Merged** in #171.
- [x] **[context, signals]** cfb28a0 discard last changes and refactor completed signals — partially **already merged**, partially **merged** in #171.
- [x] **[outercontext]** 561b202 propagate properties of component root objects to outer context — **Postponed** in #241.
- [x] **[context]** 871fbf0 Component context optimization.. maybe we do not need it? — **Merged** in #249.
- [x] **[context]** bbcf79f some bugfixes on context — **Merged** in #212.
- [x] **[fontbug]** 6934ddd small bugfix — **Rejected**, already done in b0fdbe20ad06e3424d5ef34806f517af1e19a2ad (#186).
- [x] **[optimize]** 690c82b small optimization — **Rejected**, is not worth it at the current state, optimizations would be performed at a later stage.
- [x] **[debug]** bae9d88 keep file path information in exceptions during qml parse — **Merged** in #205.
[...]

PR #168, opened 2016-04-09
failing test case

pavelvasev:
The changed signal of alias must call it's target property changed signal. Good for arrays modifications and manual changed invocation, when set is not used.

topic aliaschangeback at
https://github.com/qmlweb/qmlweb/issues/32#issuecomment-190834951

### DIFF
--- a/tests/failingTests.js
+++ b/tests/failingTests.js
@@ -19,6 +19,7 @@ window.failingTests = {
     ],
     properties: [
       'alias have changed signal',
+      'alias propagates it\'s changed signal back to referenced property',
       'alias to id with same name',
       'ChangedExpressionSignal',
       'StringConversion',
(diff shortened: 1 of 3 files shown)


tests/failingTests.js
```diff
@@ -22,7 +22,8 @@ window.failingTests = {
       'alias to id with same name',
       'ChangedExpressionSignal',
       'StringConversion',
-      'undefined property has undefined value'
+      'undefined property has undefined value',
+      'alias propagates it\'s changed signal back to referenced property'
```
ChALkeR: Could you insert this under the `'alias to id with same name',` instead?
I would reduce the affected lines count and would keep the list sorted.

ChALkeR:
LGTM with a nit.

PR #169, opened 2016-04-09
comment on signal callType internal var

pavelvasev:


ChALkeR:
/cc @akreuzkamp 

I have little idea how that code works =)

Plaristote:
Alright, so this doesn't look bad to me.

But I'd much rather have these things available as names in the code, rather than in a comment. I guess it's the C++ developer talking, but that seems to be something that should be described in an enum.

I see that these callTypes are also referenced outside of the scope of disconnect, in the isConnected function. So I think it would make sense to use an enum-like object to describe these behaviors.

ChALkeR:
Ok, this does LGTM as a temporary solution.
It's better to have a comment there than no explanations at all =).

PR #171, opened 2016-04-09
createQmlObject

pavelvasev:


Plaristote:
LGTM.

We have the same issue about that executionContext parameter with createComponent. It may be worthy of creating an issue about it, once this  is merged.

src/qtcore/qml/QMLEngine.js
```diff
@@ -127,10 +127,22 @@ QMLEngine = function (element, options) {
...
-        // Call completed signals
+        // #engine [engine::loadQMLTree] -> [engine::callCompletedSignals]
+        this.callCompletedSignals();
+    }
+
+    this.callCompletedSignals = function () {
+        // the while loop is better than for..in loop, because completedSignals array might change dynamically when
+        // some completed signal handlers will create objects dynamically via createQmlObject or Loader
+        while (this.completedSignals.length > 0) {
+           var h = this.completedSignals.splice(0,1); // remove first handler from list and put it to `h`
```
ChALkeR: Use `.shift()`.

src/qtcore/qml/QMLEngine.js
```diff
@@ -127,10 +127,22 @@ QMLEngine = function (element, options) {
...
+           h[0](); // call first handler
+        }
+        /* old way:
         for (var i in this.completedSignals) {
             this.completedSignals[i]();
         }
+        */
```
ChALkeR: No need to keep the old way, git stores the history just fine =).

src/qtcore/qml/QMLEngine.js
```diff
@@ -127,10 +127,22 @@ QMLEngine = function (element, options) {
 
         this.start();
 
-        // Call completed signals
+        // #engine [engine::loadQMLTree] -> [engine::callCompletedSignals]
+        this.callCompletedSignals();
+    }
+
+    this.callCompletedSignals = function () {
```
ChALkeR: Since this is a new function, why not create it in `QMLEngine.prototype` from the start?

tests/QMLEngine/qml/BasicCompletedOfDynamicObjects.qml
```diff
@@ -0,0 +1,10 @@
+import QtQuick 2.0
+
+Rectangle {
+  color: "red"
+  id: item
+  
+  Component.onCompleted: {
+    Qt.createQmlObject( "import QtQuick 2.2\nItem { Component.onCompleted: parent.color = 'cyan'; }", item,"inlinecode1",__executionContext );
```
ChALkeR: Why do we need a fourth parameter here? It's not documented in QML afaik.

Plaristote: It's a bummer, but we already have that issue with the createComponent method. We don't have yet something that would allow us to know which executionContext to use within createComponent and createQmlObject.

I've been trying to merge the realinheritance branch, and it seems to implement a variable that would give us the proper executionContext. But realinheritance is insanely hard to merge... some of the first commits are so large than github can't make a diff, and master has diverged so far from kde-master that if we did merge realinheritance as it is now, we'd break a lot of stuff for sure.

Plaristote:
Also, is this a replacement for `fork-pavelvasev` createqmlobject commits ?

pavelvasev:
@Plaristote yes, it is a replacement of some commits from that fork, cleaned and dressed.

@ChALkeR may I ask you to perform changes you pointed out? I'm going on vacations for 2 weeks and very limited in time now.. Thanks in advance..

BTW the code is commented and not wiped to prevent some developer in future to refactor loop in for-each of for- fashion. If you think it's ok to remove comment at all, please do it.

ChALkeR:
> BTW the code is commented and not wiped to prevent some developer in future to refactor loop in foreach or for fashion. If you think it's ok to remove comment at all, please do it.

Just the comment explaining why it's dont that way is enough.

ChALkeR:
Ok, LGTM.

This needs futher work on removing the `__executionContext` of `createQmlObject`, though. See https://github.com/qmlweb/qmlweb/pull/171#discussion_r59693671.

Plaristote:
About the `__executionContext`, the problem had been fixed in the realinheritance branch. I've merge the solution, and made an extra commit to change createQmlObject's tests so that it checked that the context had been properly acquired. See #181.

PR #128, opened 2016-03-07
@igosoft patches (split), remaining patches

ChALkeR:
This is a redo of #121, keeps a better track of remaining changes.

This is in-progress, do not review yet.
Just making a PR to make this visible.

This consists of [igosoft-patches-split](https://github.com/qmlweb/qmlweb/commits/igosoft-patches-split) branch minus the `[ignore]` and `[merged]` commits (based on top of the [igosoft-patches-merged](https://github.com/qmlweb/qmlweb/commits/igosoft-patches-merged) branch).

~~`[good]` does not mean that it's properly reviewed or ready for landing, it means that the commit was split.~~ *Update: removed `[good]` labels.*

[igosoft-patches-split](https://github.com/qmlweb/qmlweb/commits/igosoft-patches-split) branch is exactly the same as [igosoft-patches](https://github.com/qmlweb/qmlweb/commits/igosoft-patches) branch (i.e. no code difference), but with split commits.

Issue #62, opened 2016-02-21
Merge igosoft-patches


akreuzkamp:
See branch igosoft-patches.

These patches contain a Loader class, childrenRect and some small fixes.

Unfortunately everything is in one commit. Probably needs some splitting up and needs to be rebased on top of current master.
See also https://phabricator.kde.org/D623

/cc @igosoft Maybe you can set hand to it yourself, you know your patches best :)

ChALkeR:
@igosoft, do you have these separate somewhere, or was this initially one huge change?

This has to be split somehow. I can start working on that.
I will keep two additional branches — `igosoft-patches-split` (with the same changes as here and zero diff) and `igosoft-patches-split-rebased` (some of the splitted changes cherry-picked and rebased against master).

Note that there already is an initial `Loader` implementation, but this one looks better to me.
@igosoft, in the mean time, please leave a comment in #52.

ChALkeR:
https://github.com/qmlweb/qmlweb/commits/igosoft-patches — original.
https://github.com/qmlweb/qmlweb/commits/igosoft-patches-split — split.

ChALkeR:
This is a work in progress.

ChALkeR:
@igosoft I'm reverting code moving and your microoptimizations, those are mostly useless and will make merges harder.

Also, there is a mistake here, for example: https://github.com/qmlweb/qmlweb/commit/600e3bd5697aae721a1e7f5f01671719d9bdddd0#diff-de79c85e70cebf990f54694973fa1f62R11990.

But the changes look great overall, thanks! I appreciate the Loader especially =), hope to test it soon.

ChALkeR:
@igosoft, Optimizations (if needed) and code cleanup will be redone after merging everything.

ChALkeR:
Status update: the code splitting is half done, will continue soon.

igosoft:
No. I don't have them separate. At that time I was playing with qmlqweb and done many mistakes. ChALkeR please continue to merge my changes. In case of any questions just let me know.

ChALkeR:
Still working on this, about half-done splitting.

henrikrudstrom:
@igosoft, @ChALkeR would be great to get this one landed, anything i can do to help?

igosoft:
The work is in progress but it depands on @ChALkeR how and if we can help here.

ChALkeR:
See #121 (not ready yet).

ChALkeR:
[...]

PR #172, opened 2016-04-09
@igosoft patches merge — second batch 

ChALkeR:
Ref: #128, #62.

This one is small, hoping for a quick review here.

/cc @igosoft @Plaristote @akreuzkamp


src/qtcore/qml/UpdateGeometry.js
```diff
@@ -194,6 +196,38 @@ function updateVGeometry(newVal, oldVal, propName) {
...
+    var minY = children.length>0 ? children[0].y : 0;
+    var h=0;
+    var w=0;
+    var child;
+
+    for (var i=0;i<children.length; i++){
+        child = children[i];
+
+        h = child.$isUsingImplicitHeight ? child.implicitHeight : child.height;
+        w = child.$isUsingImplicitWidth ? child.implicitWidth : child.width;
```
akreuzkamp: This shouldn't be necessary, unless you check values manually using child.$properties.implicitHeight.value.
child.height already gives you it's implicitHeight, if it's using that.

ChALkeR: @akreuzkamp, it is also used in a many more places at c1bf9083d7beabf53d676ec2f187f82cde973e97.

Are all those `child.$isUsingImplicitHeight ? child.implicitHeight : child.height` and `child.$isUsingImplicitWidth ? child.implicitWidth : child.width` unnecessary?

/cc @igosoft

src/qtcore/qml/elements/QtQuick.Controls/ScrollView.js
```diff
@@ -0,0 +1,109 @@
...
+  name:     'ScrollView',
+  versions: /.*/,
+  baseClass: QMLItem,
+  constructor: function QMLScrollView(meta) {
+    QMLItem.call(this, meta);
+
+    var self = this;
+    
+ 
+    this.dom.style.pointerEvents = "auto";
```
akreuzkamp: you can use this.css ;)

src/qtcore/qml/UpdateGeometry.js
igosoft: @akreuzkamp @ChALkeR I can get rid of those "checks" if they aren't necessary. Please let me know.

akreuzkamp: If $isUsingImplicitHeight is set, implicitHeight and height are synchronized. That's the idea of implicitHeight, after all ;)
So yes, I'm pretty sure that all of these checks are unnecessary (except of setting css.height to auto, but that's something different anyway :) )

src/qtcore/qml/elements/QtQuick/MouseArea.js
```diff
@@ -81,6 +83,16 @@ registerQmlType({
...
+    this.dom.onmouseover = function(e) {
+        self.containsMouse = true;
+        self.dom.style.cursor = cursorShapeToCSS();
+        self.entered();
+    }
+    this.dom.onmouseout = function(e) {
+        self.containsMouse = false;
+        self.dom.style.cursor = "auto";
+        self.exited();
+    }
```
Plaristote: I don't understand why we have to do this. If you simply watch for the `cursorShape` property, and set the `dom.style.cursor` value whenever it changes, won't it do the same thing ? Do we really need to update the css attribute everytime the mouse enters or exits it ?

igosoft: Indeed, changing style here doesn't make sense :(. I can't remember why I did such a change :). I think we can remove these lines (self.dom.style.cursor). Do you want me to remove these lines now ?

Plaristote: I've made a commit to update the behavior. I also found an issue with the `pressedButtons` property: in an event handler, it was refered as `this.pressedButtons`. But inside the event handlers, `this` isn't the QML component: that's why we have the `self` variable around, sometimes. I've also made a commit to fix that.

ChALkeR: Yes, I missed that to. That should have been an arrow function, transplited by babel =).
But lets use `self` for now.

src/qtcore/qml/elements/QtQuick/MouseArea.js
```diff
@@ -89,5 +91,41 @@ registerQmlType({
...
+          case Qt.ClosedHandCursor: cursor = "move";break;
+          case Qt.DragCopyCursor: cursor = "auto";break;
+          case Qt.DragMoveCursor: cursor = "auto";break;
+          case Qt.DragLinkCursor: cursor = "auto";break;
+          case Qt.LastCursor: cursor = "auto";break;
+          case Qt.BitmapCursor: cursor = "auto";break;
+          case Qt.CustomCursor: cursor = "auto";  break;
+        }
+        return cursor;
+    }
```
ChALkeR: btw, this could be simplified a lot. Will add an additional commit.

ChALkeR:
@igosoft I'm going to squash your updates here.

ChALkeR:
I added 7fba33b6ea04dd7879eddaae7bf8b021699b13ad, it implements all those cursors, excluding the custom ones and the `LastCursor` (which is now an alias to DragLinkCursor).

Not sure what `LastCursor` does and where it's documented, btw.

src/qtcore/qml/UpdateGeometry.js
```diff
@@ -95,6 +95,8 @@ function updateHGeometry(newVal, oldVal, propName) {
         this.width = width;
 
     this.$updatingGeometry = false;
+
+    if (this.parent != undefined) updateChildrenRect(this.parent);
```
ChALkeR: This is the same as `if (this.parent)`. It should be either that or `if (this.parent !== undefined)`.

src/qtcore/qml/UpdateGeometry.js
```diff
@@ -194,6 +196,32 @@ function updateVGeometry(newVal, oldVal, propName) {
         this.height = height;
 
     this.$updatingGeometry = false;
+
+    if (this.parent != undefined) updateChildrenRect(this.parent);
```
ChALkeR: Ditto.

src/qtcore/qml/UpdateGeometry.js
igosoft: @ChALkeR done

ChALkeR:
LGTM

Will squash two commits a bit later, though.

ChALkeR:
Squashed, added one more commit on top of that to de-duplicate some code in ScrollView.
Everything here LGTM, waiting for more reviews.

Plaristote:
LGTM

PR #174, opened 2016-04-14
Merge the remaining arnopaehler patches

ChALkeR:
This has TextEdit updates, TextField implementation, and minor gitignore changes.

Ref: #163

### DIFF
--- a/src/qtcore/qml/elements/QtQuick/TextEdit.js
+++ b/src/qtcore/qml/elements/QtQuick/TextEdit.js
@@ -3,8 +3,89 @@ function QMLTextEdit(meta) {
 
     var self = this;
 
+    // Properties
+    createSimpleProperty('bool', this, 'activeFocusOnPress');
+    createSimpleProperty('url', this, 'baseUrl');
+    createSimpleProperty('bool', this, 'canPaste');
+    createSimpleProperty('bool', this, 'canRedo');
+    createSimpleProperty('bool', this, 'canUndo');
+    createSimpleProperty('color', this, 'color');
+    createSimpleProperty('real', this, 'contentHeight');
+    createSimpleProperty('real', this, 'contentWidth');
+    createSimpleProperty('Component', this, 'cursorDelegate');
+    createSimpleProperty('int', this, 'cursorPosition');
+    createSimpleProperty('rectangle', this, 'cursorRectangle');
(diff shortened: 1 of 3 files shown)


ChALkeR:
The question here is: are _all_ those new methods of TextEdit also present in TextArea? If not, perhaps we should split them into separate files (like it was done in @arnopaehler fork)?

src/qtcore/qml/elements/QtQuick/TextEdit.js
```diff
@@ -25,14 +105,135 @@ function QMLTextEdit(meta) {
...
+    // The function below is from QMLText
+    this.$drawItem = function(c) {
+        c.save();
+        c.font = fontCss(this.font);
+        c.fillStyle = this.color;
+        c.textAlign = "left";
+        c.textBaseline = "top";
+        c.fillText(this.text, this.left, this.top);
+        c.restore();
+    }
```
ChALkeR: I will file a pull request afterwards to remove those canvas-based rendere leftovers.

ChALkeR: PR: #175. This will have to be updated after that one gets merged. Or the other way — if this gets merged first, #175 will have to be updated.

Plaristote:
The `linkAt` method doesn't exist for TextArea. The others are present in both components.

I don't think it matter so much anyway, there's still plenty of work to do in these components before they match QML.

LGTM.

ChALkeR:
@Plaristote, I updated this to exclude stuff removed by #175. This doesn't conflict with #175 anymore.

I'm not yet sure about the new file here. I still have two questions:
 1. Why do we need `updateCss` at all?
 2. Do we want to keep it in a separate file (will we reuse it)? It's currently used only in TextEdit.

Plaristote:
Come to think of it... that depends on how we'll do in the future.

There was a commit about allowing us to chose which html tag would be used for a given component. Maybe we should use that instead of appending a textarea child to the component.
If we do that, we won't need `updateCss` anywhere. And we shouldn't keep it at all. Maybe we should re-write it this way.

ChALkeR:
@Plaristote Ok, that could be fixed later. For now, I just included `updateCss` into the TextEdit, as we likely are not going to re-use it anywhere else.

Plaristote:
Fine by me.
LGTM.

PR #175, opened 2016-04-14
Remove canvas backend leftovers

ChALkeR:
This removes the remaining `.$draw()`, `.$drawItem()`, and `.$perfDraw()` methods.
Those were not used since the canvas renderer was removed.

/cc @Plaristote, @akreuzkamp.

### DIFF
--- a/src/qtcore/qml/QMLBaseObject.js
+++ b/src/qtcore/qml/QMLBaseObject.js
@@ -4,7 +4,6 @@ function QMLBaseObject(meta) {
     var i,
         prop;
 
-    this.$draw = function(){};
     this.$isComponentRoot = meta.isComponentRoot;
     this.$context = meta.context;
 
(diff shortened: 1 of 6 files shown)


Plaristote:
I'm even surprised not to have noticed those before.
This LGTM. But we can also use this PR to find the last remaining references to the old canvas renderer.

For instance, there's a canvas variable in the QMLEngine file that gets used from time to time. I'm guessing it's useless now. And I'm guessing some functions using it might be unused as well.

ChALkeR:
@Plaristote, there are way more unused methods, but I wanted to create other pull requests for those, because some of those might or might not be usable in the future.

I missed the canvas variable, though. I will update this pull request to inlude that.

ChALkeR:
Updated to also remove `fontCss()` and `$getTextMetrics()`.

Plaristote:
That's all of them as far as I can tell.
LGTM.

ChALkeR:
@Plaristote `touchHandler`/`mousemoveHandler` are also unused, but those would be in another PR. I am not sure if we would need those or not.

PR #177, opened 2016-04-14
fix js import test not passing

Plaristote:
* replace qmlEngine references with engine
* fix jsparser not properly walking the AST_Tree
* enable the QMLEngine JavascriptImport test

I noticed that js import wasn't working. Then I noticed the test was disabled. So I found what was blocking the test and fixed it (qmlEngine was used instead of engine in some place.... but qmlEngine is defined by the AutoLoader, and it's exclusively for use outside of qmlweb's scope. So I replaced the occurences of qmlEngine with engine).

Then I fixed a bug in jsparser, not properly using the AST_Tree returned by qmlparse.

The test was wrong as well: if we set a color to 'magenta', we shouldn't expect the CSS property to be 'magenta', but the rgb value for the magenta color (as this is the new behavior for QMLColor).

### DIFF
--- a/src/qtcore/qml/QMLEngine.js
+++ b/src/qtcore/qml/QMLEngine.js
@@ -460,7 +460,7 @@ QMLEngine = function (element, options) {
         if (name in this.components)
             return this.components[name];
 
-        var file = qmlEngine.$basePath + name + ".qml";
+        var file = engine.$basePath + name + ".qml";
 
         this.ensureFileIsLoadedInQrc(file);
         tree = convertToEngine(qrc[file]);
--- a/src/qtcore/qml/elements/Component.js
+++ b/src/qtcore/qml/elements/Component.js
@@ -54,8 +54,8 @@ function QMLComponent(meta) {
         var src = importDesc[1];
         var js;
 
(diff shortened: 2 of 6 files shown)


src/qtcore/qml/QMLEngine.js
```diff
@@ -115,7 +115,6 @@ QMLEngine = function (element, options) {
 
         // Create and initialize objects
         var component = new QMLComponent({ object: tree, parent: null });
-
```
ChALkeR: Unrelated style change, let's keep those minimal until we finish with the merges..

Plaristote: Hadn't noticed that one. It's fixed.

ChALkeR:
Those parser lines are from 5be63f5d.

Everything except for the parser LGTM, and I can't give any comments about the parser — I am not familiar with that code.

src/qtcore/qml/QMLEngine.js
ChALkeR: Thanks!

ChALkeR:
On a second thought, this does LGTM. Merging.

PR #178, opened 2016-04-14
Allow to specify html tag name for component

Plaristote:
Related to https://github.com/qmlweb/qmlweb/pull/174#issuecomment-209905700
This commit from #32 should allow us to get rid of TextEdit's `updateCss` in the future.

### DIFF
--- a/src/qtcore/qml/elements/Item.js
+++ b/src/qtcore/qml/elements/Item.js
@@ -12,8 +12,16 @@ function QMLItem(meta) {
         this.dom.style.left = "0";
         this.dom.style.overflow = "hidden"; // No QML stuff should stand out the root element
     } else {
-        if (!this.dom) // Create a dom element for this item.
-            this.dom = document.createElement("div");
+        if (!this.dom) { // Create a dom element for this item.
+           /* If some Item-based component want's to hold itself in html tag different than "div",
+               it can specify desired tag name:
+               ```
+               property var htmlTagName: "fieldset"
+               ```
+            */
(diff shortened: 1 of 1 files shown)


ChALkeR:
I think that I would prefer this to have a more specific name to avoid accidental hits and to give a hint that this is internal details. Perhaps something like `_domTagName` would do.

Also, as for `TextEdit`, we could have made that internal, and pass it as an option to `registerQmlType`, for example.

This commit allows to specify that from QML itself, and to create new such elements using only QML. That could have been done by specifying a special DomElement type (e.g. in `QmlWeb.Dom` module or however we would call it) and adding that new proprerty to it alone.

Not sure yet.

ChALkeR:
Ah, and about overriding the type of the element that we create using the internal QmlWeb API (like `TextEdit`), we already have that — take a look at [QtQuick.Controls/Button](https://github.com/qmlweb/qmlweb/blob/master/src/qtcore/qml/elements/QtQuick.Controls/Button.js#L2) and [QtQuick.Controls/Checkbox](https://github.com/qmlweb/qmlweb/blob/master/src/qtcore/qml/elements/QtQuick.Controls/Checkbox.js#L7).

Plaristote:
That make sense. I'm closing this, should we reject the commit then ?

ChALkeR:
@Plaristote I think so. This functionality should be moved to a special element, e.g. in `QmlWeb.Dom` — see #182.

Also note that `meta.object.somePropertyName.value` doesn't work in the current master, it's just `meta.object.somePropertyName`.

PR #179, opened 2016-04-14
remove unused file process.js

Plaristote:
While looking at our coverage, I found this file that was not covered at all. I didn't know what it was used for, and after checking, it turns out it's not used... at all.

After removing it, the test still pass (obviously since they didn't cover it), and my own testing apps work the same way as well.
I've looked thoroughly, and haven't found anything in our code that references this file.

I'm not entirely sure what this file was used for. Lpaimen was the only one to have worked with this file, or so it seems.


ChALkeR:
Don't forget to remove it from `gulpfile.js`.

ChALkeR:
That file was introduced in f7402cf8fd8d7ce3366461fb1f932714f82b2066 by @lpaimen.

ChALkeR:
And perhaps not used anymore after 2cc766fa.

/cc @akreuzkamp

ChALkeR:
LGTM, but I would like to hear from @akreuzkamp here =).

ChALkeR:
This reduces the total lines count of the library by 20%, btw. =)

akreuzkamp:
I'm like 99.973% sure, it's not used and never has been, since I work on qmlweb :) Let's finally get rid of it :)

PR #181, opened 2016-04-14
Merge kde realinheritance

Plaristote:
These are the progresses I'm making merging the realinheritance branch.
I'm trying to split the major overhaul into smaller commits, while making sure not to break anything after each step.

Right now I have the first commit, and a small bit of the second one, which consisted in changing the prototype for createSimpleProperty, plus renaming that function to createProperty.

This is not especially up for a merge yet. I'm just opening this so that we can discuss the changes, and whether this merge is following the right direction.


src/qtcore/qml/elements/QtQuick/Row.js
```diff
@@ -9,9 +9,9 @@ registerQmlType({
 function QMLRow(meta) {
     QMLPositioner.call(this, meta);
 
-    createProperty({ type: "enum", object: this, name: "layoutDirection" });
+    createProperty({ type: "enum", object: this, name: "layoutDirection", initialValue: 0 });
     this.layoutDirectionChanged.connect(this, this.layoutChildren);
-    this.layoutDirection = 0;
+    this.layoutDirectionChanged(0);
```
akreuzkamp: why do we need this line? Isn't this.layoutChildren(); rather what we want?

src/qtcore/qml/QMLPositioner.js
```diff
@@ -1,12 +1,12 @@
...
     QMLItem.call(this, meta);
 
-    createProperty({ type: "int", object: this, name: "spacing" });
+    createProperty({ type: "int", object: this, name: "spacing", initalValue: 0 });
     this.spacingChanged.connect(this, this.layoutChildren);
     this.childrenChanged.connect(this, this.layoutChildren);
     this.childrenChanged.connect(this, QMLPositioner.slotChildrenChanged);
 
-    this.spacing = 0;
+    this.spacingChanged(0);
```
akreuzkamp: why do we need this line? Isn't this.layoutChildren(); rather what we want?

akreuzkamp:
looks (almost) good to me :)

src/qtcore/qml/elements/QtQuick/Row.js
Plaristote: I'm trying to preserve the original behavior as best as I can.
At some places, the default value for the property were only set after connecting functions to the property's changed signal. In those cases, I've made sure the signals were still going to be triggered.
I'm not sure it does something in every cases : but I know that at least in some, it is important (for instance, I forgot it at some point on the Image item, and it broke an image test).

src/qtcore/qml/QMLPositioner.js
Plaristote: Actually, forget my previous reaction : you're right, this.layoutChildren is more appropriate. I've fixed it.

ChALkeR:
I'm not exactly sure about changing `createProperty` to receive all parameters inside an options object.

IMO it would be better to keep `type`/`object`/`name` arguments as they are now in master, and put optional arguments (`initalValue`) inside an options array, like this:
```js
createProperty("int", this, "spacing", { initalValue: 0 });
```

That would keep the code a little shorter and easier to write.

ChALkeR:
About the `initialValue` for `string`/`list` — shouldn't that always be set to a valid string or list? If yes, shouldn't it default to `""` and `[]` inside the `createProperty` itself? I'm not about how Qt QML behaves yet.

ChALkeR:
These commits LGTM:
 * «Fix some closure compiler warnings.»
 * «stores the current execution context in a variable»
 * «remove extra params from Qt.createQmlObject and Qt.createComponent»

akreuzkamp:
Yes, properties could and should be default constructed automatically, nowadays. It's a matter of calling this.value = constructors[this.type](); inside the QMLProperty constructor (and maybe fixing some basic type constructors).

I'm open to either having createProperty arguments as one object or multiple arguments. :)

ChALkeR:
@Plaristote Could you please split this into two pull requests, one with the `createProperty` changes (two commits) and the one with everything else, so we could discuss those separately? They should be independent, as it looks to me.

Plaristote:
See #186 for that.

ChALkeR:
I'm working on an alternate version of this.

ChALkeR:
Update: see #200.

PR #183, opened 2016-04-18
Process sources with Babel and remove encapsulate helper

ChALkeR:
These are two independant changes.

[gulp-iife](https://www.npmjs.com/package/gulp-iife) is used instead of the [encapsulate](https://github.com/qmlweb/qmlweb/blob/master/src/helpers/encapsulate.begin.js) [helper](https://github.com/qmlweb/qmlweb/blob/master/src/helpers/encapsulate.end.js).

[Babel](http://babeljs.io/) allows us to use more recent ECMAScript features under current (and older) browsers.


akreuzkamp:
why replace use strict?

ChALkeR:
@akreuzkamp Because we don't support it yet, I have no idea how most of that code would work under `use strict` (we need to run eslint first), and some of that code is definitely broken under `use strict`, e.g. `with(…)`.

akreuzkamp:
ah, Babel adds "use strict"?

ChALkeR:
@akreuzkamp To be clear: we don't currently have `"use strict";` in our code, Babel adds it with that preset.

akreuzkamp:
Ok. Lgtm, then :)

PR #184, opened 2016-04-18
Move orphaned elements to corresponding modules

ChALkeR:
This moves Settings to Qt.labs.settings and RestModel to QmlWeb.


Plaristote:
Good initiative, LGTM.

PR #185, opened 2016-04-18
@igosoft merge: QMLPositioner: fix layoutChildren triggering

ChALkeR:
> QMLPositioner: fix layoutChildren triggering
> 
> In QML opacity doesn't change children layout.

Not sure about the additions here — do we need to update layout on implicit size changes?

/cc @igosoft @akreuzkamp @Plaristote

Ref: #128, #62.

### DIFF
--- a/src/qtcore/qml/QMLPositioner.js
+++ b/src/qtcore/qml/QMLPositioner.js
@@ -17,10 +17,12 @@ QMLPositioner.slotChildrenChanged = function() {
             child.widthChanged.connect(this, this.layoutChildren);
         if (!child.heightChanged.isConnected(this, this.layoutChildren))
             child.heightChanged.connect(this, this.layoutChildren);
+        if (!child.implicitWidthChanged.isConnected(this, this.layoutChildren))
+            child.implicitWidthChanged.connect(this, this.layoutChildren);
+        if (!child.implicitHeightChanged.isConnected(this, this.layoutChildren))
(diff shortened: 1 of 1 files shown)


Plaristote:
After reading the code, I think it's important to keep layoutChildren on the opacity event. Although we might wanna check QML's  behavior on this.

Right now, when a child's opacity reaches nil, the element disappears altogether, which is why the layout needs to computed again. Whatever QML does (making the element invisible, but still taken in accounts for layout positioning, or making it disappear as it does right now), we can't accept this as is.

ChALkeR:
@Plaristote That looks like a bug to me. See #191.

igosoft:
I can be wrong but opacity can't change a position or existence of component. It should to control only the visibility of a component.

here is my test (Qt5)

import QtQuick 2.0
 Row {
   width: 300
   height: 200

   Column {
     height: parent.height
     width: 100
     Rectangle {
         id: rect
       height: 100
       width: parent.width
       color: 'red'
       opacity: 0
     }
     Rectangle {
       height: 100
       width: parent.width
       color: 'pink'
       MouseArea
       {
           id: ma
           anchors.fill: parent
           onClicked: rect.opacity = !rect.opacity
       }
     }
   }
 }

ChALkeR:
This is blocked by the opacity bug.

ChALkeR:
Superseeded by #217.

PR #188, opened 2016-04-18
Delay module registation until the base class is loaded.

ChALkeR:
Implement string-based baseClass specification.
This allows to delay module registration until the base class is loaded.

This fixes the issue discussed here: https://github.com/qmlweb/qmlweb/pull/160#issuecomment-206384442 in a sane way, without forcing us to manually specify the element order in the gulpfile.

It currently also supports a short syntax, i.e. `Repeater` instead of `QtQuick.Repeater` when used from `QtQuick` module. Is it worth it, or does that just overcomplicate things?

Not all modules were converted to the new syntax yet, only those ones that failed without it.

Also this PR adds QmlWeb.Basic module, which I propose would include supportive classes to implement QtQuick modules, but which should not be present in QtQuick itself.

This would also be required to split modules into separate files (one file per module), see https://github.com/qmlweb/qmlweb/issues/143#issuecomment-210131659. Cross-dependencies between modules would otherwise prevent that, and we have cross dependencies at least between QtQuick module and the helper classes that are not part of the module itself (moved to QmlWeb.Basic here).

/cc @Plaristote, @akreuzkamp. I would like to see your reviews here.

### DIFF
--- a/src/qtcore/qml/qml.js
+++ b/src/qtcore/qml/qml.js
@@ -24,6 +24,8 @@ var modules = {
     Main: constructors
   };
 
+const dependants = {};
+
 // Helper. Adds a type to the constructor list
 global.registerGlobalQmlType = function (name, type) {
   global[type.name]  = type;
@@ -35,7 +37,36 @@ global.registerGlobalQmlType = function (name, type) {
 global.registerQmlType = function(options) {
   if (typeof options != 'object') {
     registerGlobalQmlType(arguments[0], arguments[1]);
-  } else {
+    return;
+  }
+
+  if (typeof options.baseClass === 'string') {
+    // TODO: Does not support version specification (yet?)
+    var baseModule, baseName;
+    const dot = options.baseClass.lastIndexOf('.');
+    if (dot === -1) {
+      baseModule = options.module;
+      baseName = options.baseClass;
+    } else {
+      baseModule = options.baseClass.substring(0, dot);
+      baseName = options.baseClass.substring(dot + 1);
+    }
+    const found = (modules[baseModule] || [])
+                    .filter(descr => descr.name === baseName);
+    if (found.length > 0) {
+      // Ok, we found our base class
+      options.baseClass = found[0].constructor;
+    } else {
+      // Base class not found, delay the loading
+      const baseId = [baseModule, baseName].join('.');
+      if (!dependants.hasOwnProperty(baseId)) {
+        dependants[baseId] = [];
+      }
+      dependants[baseId].push(options);
+      return;
+    }
+  }
+
     var moduleDescriptor = {
       name:        options.name,
       versions:    options.versions,
@@ -49,6 +80,11 @@ global.registerQmlType = function(options) {
     if (typeof options.baseClass !== 'undefined') {
       inherit(options.constructor, options.baseClass);
     }
+
+  const id = [options.module, options.name].join('.');
+  if (dependants.hasOwnProperty(id)) {
+    dependants[id].forEach(opt => global.registerQmlType(opt));
+    dependants[id].length = 0;
   }
 };
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -19,7 +19,7 @@ var qtcoreSources = [
   'src/qtcore/qml/qml.js',
   'src/qtcore/qml/QMLBaseObject.js',
   'src/qtcore/qml/elements/Item.js',
-  'src/qtcore/qml/QMLPositioner.js',
+  'src/qtcore/qml/elements/QtQuick/Positioner.js',
   'src/qtcore/qml/elements/QtQuick/Repeater.js',
   'src/qtcore/qml/**/*.js'
 ];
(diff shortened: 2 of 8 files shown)


ChALkeR:
Also this PR creates a QmlWeb.Basic module, which I propose would include supportive classes to implement QtQuick modules, but which should not be present in QtQuick itself.

ChALkeR:
It currently support a short syntax, i.e. `Repeater` instead of `QtQuick.Repeater` when used from `QtQuick` module. Is it worth it, or does that just overcomplicates things?

Plaristote:
LGTM

IMO the short syntax is over-complicating things, for little gain. We're going to have so much more modules, it doesn't make much sense to me to make this shortcut only for QtQuick.

I agree about the QmlWeb.Basic module too, it might be a good way to make the QML object code more uniform than it is right now.

In the future, if we separate the modules from the core of the application, we should also have a way to tell the user that some objects couldn't be loaded, because their parent object are part of a module that was never imported... (although we might also implement something to import the module on our own, and then warning the user if the module couldn't be imported... point is, this code will have to be patched if we move to the one-module/one-repository model, right ?)

ChALkeR:
> IMO the short syntax is over-complicating things, for little gain. We're going to have so much more modules, it doesn't make much sense to me to make this shortcut only for QtQuick.

The shortcut is not for `QtQuick`, but for the current module. Inside module `A`, elements could be addressed as `Element` instead of `A.Element`.

ChALkeR:
> In the future, if we separate the modules from the core of the application, we should also have a way to tell the user that some objects couldn't be loaded, because their parent object are part of a module that was never imported... (although we might also implement something to import the module on our own, and then warning the user if the module couldn't be imported... point is, this code will have to be patched if we move to the one-module/one-repository model, right ?)

Yes, that would be possible, but I was planning to do that after the initial module files loading implementation.

Plaristote:
Oops, sorry, I misunderstood. If the short syntax is for the current module, then it definitely sounds useful.

ChALkeR:
Perhaps it should be `QmlWeb.Basic/Positioner` instead of `QmlWeb.Basic.Positioner`? That would simplify some code a bit and does look more clear to me.

src/qtcore/qml/qml.js
```diff
@@ -49,7 +80,9 @@ global.registerQmlType = function(options) {
     if (typeof options.baseClass !== 'undefined') {
       inherit(options.constructor, options.baseClass);
     }
-  }
+
+  const id = [options.module, options.name].join('.');
+  (dependants[id] || []).forEach(opt => global.registerQmlType(opt));
```
ChALkeR: Does not clean up `dependants[id]`. Will add `dependants[id] = null;` here.

src/qtcore/qml/qml.js
```diff
@@ -35,7 +37,36 @@ global.registerGlobalQmlType = function (name, type) {
...
+      baseName = options.baseClass.substring(dot + 1);
+    }
+    const found = (modules[baseModule] || [])
+                    .filter(descr => descr.name === baseName);
+    if (found.length > 0) {
+      // Ok, we found our base class
+      options.baseClass = found[0].constructor;
+    } else {
+      // Base class not found, delay the loading
+      const baseId = [baseModule, baseName].join('.');
```
akreuzkamp: why do you rejoin the two names, instead of just using options.baseClass?

ChALkeR: Because it could be a short variant, e.g. `'Item'` when used from `QtQuick` module.

akreuzkamp:
Why do you put Positioner into a Basic module? Shouldn't it be a part of QtQuick? After all it's UI-related.
In the future, we'd need at least a way to automatically import QtQml when QtQuick is imported, like Qt does it (if you neither import QtQuick nor QtQml, you can't even use QtObject ;) )
Code wise it looks good to me :)

ChALkeR:
@akreuzkamp Because it's not the same `Positioner` as in QtQuick.
http://doc.qt.io/qt-5/qml-qtquick-positioner.html is a different thing.

`ListView`/`Grid`/`Row` etc inherit directly from `Item` in QtQuick.

akreuzkamp:
But our Positioner is exactly the place, where you would add these attached properties. If we want to emphasize the difference between them (which I see no need to, because http://doc.qt.io/qt-5/qml-qtquick-positioner.html only adds attached properties), shouldn't we rather change the name, than moving it to a different module?

ChALkeR:
@akreuzkamp But if we change its name, we would add an element to QtQuick which is not present there in Qt. This is why I wanted to create a separate module for things like that.

I could move that file to QtQuick instead for now, and revisit this later, if you wish =).

What about `Module.Something/Element` vs `Module.Something.Element` (where «`Module.Something`» is the module name)? Which would you prefer?

akreuzkamp:
I don't think keeping the name is a problem. After all Qt's Positioner holds (attached) properties that ought to be present in Row, Column, Grid, ..., so it's consistent with our Positioner, isn't it ?

I don't have a specific preference. Pro dot syntax is that it's consistent with the syntax inside QML 
(If we want to add version numbers here, that would break QML style syntax anyway, though), slash syntax has the advantage of code simplicity. Do as you prefer :)

akreuzkamp:
lgtm :)

PR #189, opened 2016-04-18
tests: Fix an eslint long line warning

ChALkeR:
This got missed in #171.

Perhaps we should make lint errors real errors on Travis, not just warnings.


Plaristote:
LGTM,
And yes, we should probably do that.

PR #194, opened 2016-04-18
@pavelvasev merge: Support importing code from GitHub

ChALkeR:
This merges cc54d05 and 8cea5b6.

Note: I have not reviewed this code yet, and I am not yet sure if we want it.

Ref: #32.

### DIFF
--- a/src/qtcore/qml/QMLEngine.js
+++ b/src/qtcore/qml/QMLEngine.js
@@ -88,7 +88,18 @@ QMLEngine = function (element, options) {
 
     this.extractBasePath = function( file ) {
        var basePath = file.split(/[\/\\]/); // work both in url ("/") and windows ("\", from file://d:\test\) notation
-       basePath[basePath.length - 1] = "";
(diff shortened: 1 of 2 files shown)


Plaristote:
It's my opinion, but I don't want this. Doesn't sound like something that we should provide ourselves in qmlweb.

ChALkeR:
Perhaps we could replace this with some abstract solution that would support user-supplied path processors for those two methods?

ChALkeR:
I think that I also don't like this in the current state.

Plaristote:
Sounds like the way to go.

We're most likely going to rewrite the better part of the import code anyhow. Perhaps we should just add this to the todo-list for this task (I see we haven't opened an issue on it... but I'm pretty sure I read some of us talking about the import rewrite, we should probably gather what was said there and open an issue... I'll try to find it).

ChALkeR:
@Plaristote I opened #203 for that.

PR #195, opened 2016-04-18
@pavelvasev merge: Aliases now emit changes signals

ChALkeR:
This merges 780450b635332ebd5182526b0123ad297c50fff1 and the aliases part of dce7d96.

Note: I have not reviewed this code.
Note: This fixes two testcases. It also blocks sorting out the two remaining changestets of #32, so it would be better to review this quickly =).

Ref: #32.

### DIFF
--- a/src/qtcore/qml/qml.js
+++ b/src/qtcore/qml/qml.js
@@ -374,7 +374,26 @@ function applyProperties(metaObject, item, objectScope, componentScope) {
                       if (!targetProp) {
                         console.error("qtcore: target property [",prop.val.objectName,"].",prop.val.propertyName," not found for alias ",prop.name );
                       } else {
-                        targetProp.changed.connect( prop.changed );
+                        // targetProp.changed.connect( prop.changed );
+                        // it is not sufficient to connect to `changed` of source property
+                        // we have to propagate own changed to it too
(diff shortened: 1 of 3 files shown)


ChALkeR:
This fixes a testcase.

ChALkeR:
Updated, a second aliases-related commit added, a second testcase is now fixed by this changeset :wink:.

src/qtcore/qml/qml.js
```diff
@@ -360,6 +363,43 @@ function applyProperties(metaObject, item, objectScope, componentScope) {
...
+                              if (loopWatchdog) return; loopWatchdog = true;
+                              prop.changed.apply( item,arguments );
+                              loopWatchdog = false;
+                          } );
+                          prop.changed.connect( obj, function() {
+                              if (loopWatchdog) return; loopWatchdog = true;
+                              targetProp.changed.apply( obj, arguments );
+                              loopWatchdog = false;
+                          } );
+                        } ) ();
```
ChALkeR: Note: code style is not perfect here, e.g. `(function() {` wrap is not needed. But we can fix that later at the cleanup stage, I would want to keep it like this until we complete the merges.

ChALkeR:
This does LGTM, the change here is additive.

PR #196, opened 2016-04-18
@pavelvasev merge: QObject: Add .destroy() method

ChALkeR:
Extracted from 4c3bcc649447c07401a53677b801c937d03779bb.

Note: I have not reviewed this code.
Note: `.destroy()` is used in RestModel for some reason.

Ref: #32.


Plaristote:
LGTM.

I don't remember what that destroy call is supposed to do in RestModel...

PR #200, opened 2016-04-19
Merge kde realinheritance (alternate)

ChALkeR:
This is a modified version of #181.

@Plaristote, @akreuzkamp, ptal.


ChALkeR:
Argh, I got tests failing. Will fix soon =).

ChALkeR:
Fixed.

ChALkeR:
Urgh, more tests failing.

src/qtcore/qml/elements/Item.js
```diff
@@ -334,25 +333,10 @@ function QMLItem(meta) {
...
-    this.transitions = [];
-    this.state = "";
-    this.transform = [];
-    this.rotation = 0;
-    this.scale = 1;
-    this.childrenRect.x = 0;
-    this.childrenRect.y = 0;
-    this.childrenRect.width = 0;
-    this.childrenRect.height = 0;
+    this.xChanged(0);
```
akreuzkamp: Why do we want that? I guess we should call the function we need to be called directly, because the xChanged signal isn't supposed to be called from here, anyway.

src/qtcore/qml/elements/QtQuick/Flow.js
```diff
@@ -6,14 +6,13 @@ function QMLFlow(meta) {
...
+    createProperty("enum", this, "flow", {default: this.Flow.LeftToRight});
+    createProperty("enum", this, "layoutDirection", {default: 0});
+
     this.flowChanged.connect(this, this.layoutChildren);
     this.layoutDirectionChanged.connect(this, this.layoutChildren);
     this.widthChanged.connect(this, this.layoutChildren);
-
-    this.flow = this.Flow.LeftToRight;
-    this.layoutDirection = 0;
+    this.flowChanged();
```
akreuzkamp: as earlier (and as in #181) we should call this.layoutChildren directly, here.

src/qtcore/qml/qml.js
```diff
@@ -209,23 +209,32 @@ function construct(meta) {
...
     var getter, setter;
 
-    if (typeof options == 'undefined')
-      options = {};
-    else if (typeof options != 'object')
-      options = { default: options }
-
     obj[propName + "Changed"] = prop.changed;
     obj.$properties[propName] = prop;
+    if (!options.hasOwnProperty('default')) {
```
akreuzkamp: I think we should use `constructors[prop.type]();` (default constructor) for setting the default value. Either using `obj.$properties[propName].val = constructors[prop.type]();` or just using `obj.$properties[propName].set(options.default)` (if it's not set `options.default` is undefined, thus the constructor would return a default constructed variable. (In any case, we'd need to make sure that the constructor does that)

ChALkeR: Makes sense, will take a look at that, thanks!

src/qtcore/qml/elements/Item.js
ChALkeR: Thanks, I will fix that!

ChALkeR:
Added a fix to failing tests — `anchors` property types were incorrect.

src/qtcore/qml/qml.js
ChALkeR: @akreuzkamp Does not seem to be possible in a generic way. E.g. when the property is of type `Item`, the default value should be `null`, not `Item()`.

ChALkeR:
@akreuzkamp PTAL again =). I would like to keep the `typeDefaultValues` for now, though (see https://github.com/qmlweb/qmlweb/pull/200#discussion_r60258555).

src/qtcore/qml/qml.js
akreuzkamp: maybe we can go for the `obj.$properties[propName].set(options.default)` variant, in there we differentiate between those cases anyway, in case of QtObjects we can set it to `null` and else to a default constructed value (it even makes some sense to set the value to null if assigned undefined from js, although Qt throws an error in that case)

Btw. the call to set should contain a reason, but that might be unrelated.

ChALkeR: Ah, I see what you mean. Will take a look at that.

Plaristote:
Quick remark : I'm not sure 'default' is a good name for the default value. Even if in this context, we can use it, it still is a keyword, and I think some linters may take offense to that (for instance, there was a function named 'function' that had to be renamed... I don't know if the same rule applies to 'default').

ChALkeR:
@Plaristote We could take care of that with [babel-transform-es3-member-expression-literals](http://babeljs.io/docs/plugins/transform-es3-member-expression-literals/) and [transform-es3-property-literals](http://babeljs.io/docs/plugins/transform-es3-property-literals/) at the build stage, thought that would affect only the built code.

That's valid from ECMAScript 3 (1999), though.

The recent `function` → `isFunction` flag renaming was good because that's actually more descriptive — it was a flag.

On the other hand, we could change it to something else. What would be the alternative name?

ChALkeR:
I updated the above comment.

akreuzkamp:
We used to use initialValue. I think that's more descriptive than default, too, because default is also a (used) keyword in QML, stating that object definitions which aren't part of a property assignment should be assigned to that property (i.e. data is the default property of Item and delegate is the default property of Repeater)

src/qtcore/qml/qml.js
ChALkeR: @akreuzkamp That would need a separate function (or reason), imo, because manually setting an `int` to `undefined` is not allowed in QML and it shouldn't set it to `0`.

ChALkeR:
@akreuzkamp Noted. Ok, I will stick with `initialValue`.

ChALkeR:
Updated. I switched to `initialValue` and moved the `typeInitialValues` to `QMLProperty.prototype.set`.

ChALkeR:
Added a fix to one testcase, now properties declared from QML without values `property proptype propname` default to a correct value, including `undefined` for `var`.

ChALkeR:
@akreuzkamp, @Plaristote Ping =).

Plaristote:
LGTM

ChALkeR:
Added `AnimatedImage` fixes and removed its' test from the blacklist, it now passes.

Plaristote:
LGTM

ChALkeR:
@akreuzkamp could you also take a look, please?

src/qtcore/qml/qml.js
ChALkeR: @akreuzkamp This could be further improved (even with getting rid of `typeInitialValues`) using some checks like `constructors[newVal.$class].prototype instanceof QMLBaseObject`, but I'm not sure that the other parts there would work fine with that.

We can furhter improve that later, `typeInitialValues` looks like a viable solution for now to me.

Issue #52, opened 2016-02-19
Let's discuss the license


ChALkeR:
This is aimed to replace #12.

Current issues:
 1. The LICENSE is a modified BSD-3-Clause, we should pick an SPDX-approved variant.
 2. The LICENSE is a BSD-3-Clause, but the files in `src/qtcore/qml/lib/` have BSD-2-Clause headers.
 3. We bundle `uglify-js` which is `BSD-2-Clause`.
 4. Most of web js libraries just use MIT (see #12 for previous discussion).
 5. Given the 1, 2 and 3, I don't know what to put into the `license` field of `package.json`, and `npm` spits on me with these messages:

 ```
npm WARN qmlweb@0.0.4 license should be a valid SPDX license expression
 ```

I guess everyone should be fine with either of these three licenses, but I must ask everyone.

We can also discuss the preferred license here. After that, I will make a PR with the chosen license that everyone would need to review again.

Author | [BSD-3-Clause](https://spdx.org/licenses/BSD-3-Clause.html) | [BSD-2-Clause](https://spdx.org/licenses/BSD-2-Clause.html) | [MIT](https://spdx.org/licenses/MIT.html) | Preferred | Notes
------------ | ------------- | ------------- | ------------- | ------------- | ------------- | -------------
@lpaimen | + | + | + | No preference |  |
@akreuzkamp | + | + | + | BSD-2-Clause (with parser included), BSD-2-Clause or MIT (with parser excluded) |  |
@jangmarker | + | + | + |  | One commit, through KDE reviewboard |
@Somsubhra | + | + | + | No preference |  |
@JoshuaKolden | + | + | + | MIT | Only Readme changes |
@ChALkeR | + | + | + | MIT or BSD-2-Clause |  |
@Plaristote |  | + | + | MIT? |  |
@labsin | + | + | + | No preference |  |
@arnopaehler | + | + | + |  |  |
@pavelvasev | + | + | + | No preference |  |
@henrikrudstrom | + | + | + | No preference |  |
@aspidites | + | + | + | MIT | Only Readme changes |
@shry15harsh | + | + | + | No preference | One commit, through KDE reviewboard |
@igosoft | + | + | + |MIT or BSD-2-Clause |  |
@aspotashev | + | + | + | | One commit, through KDE Phabricator |
@stephenmdangelo | + | + | + | | |

Please leave a comment about yourself — are you fine with:
[...]

PR #201, opened 2016-04-20
Relicense under MIT

ChALkeR:
See #52.

I propose to do this asap, because any forks (we already have several new ones) would inherit the license.

/cc @akreuzkamp, @lpaimen, @pavelvasev for the first round of review (you are directly mentioned in the license).

`import.js` and `qmlstructure.js` go from BSD-2-Clause to MIT, everything else goes from modified BSD-3-Clause to MIT.

Parser has been already moved to a separate repo and is licensed under BSD-2-Clause.

I am not a lawyer, this kind of merging licenses is based on what I have seen elsewhere.

Another option would be to just write `Copyright (c) 2011-2016 QmlWeb contributors`, but I'm not exactly sure if that would be a correct change.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,30 +1,26 @@
-QMLWEB project is released under the "new" BSD license. As there was no
-organization behind the project, I stripped out related parts of license text.
-It's a lawyer talk, I hope I got it right.
+QmlWeb is licensed under the MIT license, as follows:
 
 """
-Copyright (c) 2012, Lauri Paimen
-Copyright (c) 2013, Anton Kreuzkamp
-All rights reserved.
+MIT License
 
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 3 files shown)


ChALkeR:
Another option would be to just write «Copyright 2011-2016 QmlWeb contributors», but I'm not exactly sure if that would be a correct change.

ChALkeR:
Rebased.

ChALkeR:
Ok, it looks it would be unreasonably hard to get everyone's consent again here.
I'm going to just information from https://github.com/qmlweb/qmlweb/issues/52 then, and not move the copyrights for those files.

ChALkeR:
Ok. This doesn't move the copyrights information anymore, so I had to add the license into `import.js` and `qmlstructure.js`. I will open a separate pull request for copyrights stuff.

I believe that this is good to merge, given that everyone agreed to this change in #52.
Going to merge now.

PR #205, opened 2016-04-21
@pavelvasev merge: keep file path information in exceptions during qml parse

ChALkeR:
This is a cherry-pick of bae9d88, ref: #32.

/cc @pavelvasev


ChALkeR:
This could be moved deeper, e.g. in `qmlweb_parse`, and this could be made not a global variable…
But we can do all that later.

LGTM as a temporary solution.

PR #209, opened 2016-04-24
tests: Do not forget 'import QtQuick'

ChALkeR:
This is a trivial change.

Though, this is needed to fix Item placement and inheritance and for further fixes in #200.

Hoping for a quick review here.


ChALkeR:
On the other hand, this is trivial, and I'm confident in it.
Going to just merge this without wasting everyones' time.

PR #210, opened 2016-04-24
AnimatedImage: fix a mistype in constructor

ChALkeR:
This is trivial, hoping for a quick review here.

This does not fix the `AnimatedImage` testcase just yet,  #200 (with more fixes) is required for that.


ChALkeR:
On the other hand, this is trivial, and I'm confident in it.
Going to just merge this without wasting everyones' time.

PR #211, opened 2016-04-24
tests: Add more initialize tests

ChALkeR:
This just adds more basic initialization tests for the elements that we have.

5% more (basic) coverage.


Plaristote:
LGTM !

PR #212, opened 2016-04-24
some bugfixes on context

Plaristote:
Going over pavelvasev's branch, I'm trying to merge the [context] category.

There's one thing I'm not sure about. The setupGetterSetter change: I do not know what it does, if you can set me up to speed with that @pavelvasev ?

### DIFF
--- a/src/qtcore/qt.js
+++ b/src/qtcore/qt.js
@@ -20,6 +20,7 @@ global.Qt = {
     // Do not perform path lookups if name starts with @ sign.
     // This is used when we load components from qmldir files
     // because in that case we do not need any lookups.
+    var origName = name;
     if (name.length > 0 && name[0] == "@") {
       nameIsUrl = true;
       name = name.substr( 1,name.length-1 );
@@ -55,7 +56,7 @@ global.Qt = {
 
     engine.loadImports( tree.$imports,component.$basePath );
 
-    engine.components[name] = component;
+    engine.components[origName] = component;
     return component;
   },
 
--- a/src/qtcore/qml/qml.js
+++ b/src/qtcore/qml/qml.js
@@ -208,7 +208,7 @@ function construct(meta) {
 
     // id
     if (meta.object.id)
-        meta.context[meta.object.id] = item;
+        setupGetterSetter(meta.context, meta.object.id, function() { return item; }, function() {});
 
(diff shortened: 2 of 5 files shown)


ChALkeR:
I'm not exactly sure about the `origName` change — does it make sense to cache using the non-normalized name?

An alternate solution would be to cache by the normalized name, i.e. moving the `if (name in engine.components) {…` block _after_ the `name` normalization.

ChALkeR:
Btw, this merges bbcf79f.

src/qtcore/qt.js
```diff
@@ -13,13 +13,18 @@ global.Qt = {
   // Load file, parse and construct as Component (.qml)
   createComponent: function(name) {
     if (name in engine.components)
-        return engine.components[name];
+    {
+      var newComponentInstance = Object.create( engine.components[name] );
+      newComponentInstance.$context = _executionContext;
+      return newComponentInstance;
+    }
```
ChALkeR: This was revered in 871fbf03c8dcdda039c5556add2bdfb1b2523c41, see second comment there.

ChALkeR:
Ok, I now undertand the `setupGetterSetter` change.

Original description:
```
        // we have to call `delete` before setting new value to context
        // because it may contain some object property with defined setter
        // in that case our assign will invoke that setter, and thus we will break some external object property!

        /// delete meta.context[meta.object.id]; 
        /// meta.context[meta.object.id] = item;

        // PROBLEM
        // we cannot just delete value from context, because it may contain setters/getters from prototype context's
        // which are not deleted by `delete`

        // SOLUTION
        // we declare new getter and empty setter. seems it is ok.
```

That means that if `context` already has a property named `meta.object.id`, then assigning to it could call the setter of that property, but we need just to override it in the context visibility. Ideally, all additions to on `context` should go though a `setupGetterSetter` then.

I would like to see a testcase when this fails without a `setupGetterSetter`, though.

ChALkeR:
Ok, added a testcase for the `setupGetterSetter` change. That change LGTM.

ChALkeR:
LGTM, merging.

ChALkeR:
Note: I also removed the change that was reverted in 871fbf0.

@Plaristote, @pavelvasev thanks!

PR #213, opened 2016-04-24
tests: Simplify initialization tests

ChALkeR:
Dependencies are now optional (and working).
They were broken before, in `module._depends` `module` was actually a string.

Version is now included in the key. It's now possible to specify different sets of initialization tests per module per version.

And no more `_version`/`_depends` required!


Plaristote:
LGTM

PR #214, opened 2016-04-27
Remove outer eval in jsparser. Babelify everything but parser.js

ChALkeR:
The main change here is 00db7e0ede1ac09a796a7ca60904ae56287b8f55, the other two are minor fixes.

`Qt.Key_Yes` was defined twice, I removed a duplicate.

/cc @Plaristote


Plaristote:
LGTM

Issue #173, opened 2016-04-09
Move parser to a separate repo


ChALkeR:
This is blocked by the merges atm.

The proposal is to move our parser to [qmlweb/qmlweb-parser](https://github.com/qmlweb/qmlweb-parser) and have that as an (optional, given gulp-qmlweb) runtime dependency for qmlweb, and as a sole qmlweb dependency for gulp-qmlweb.

qmlweb-parser would be BSD-2-Clause licensed (inherited from uglifyjs).

Refs: #143, #52.

The questions here are:
 1. Which files are we going to move?
 2. How would be the move performed?

My thoughts:
 1. I suppose that we are going to move `src/qtcore/qml/QMLBinding.js`, `src/qtcore/qml/lib/parser.js`, `src/qtcore/qml/lib/jsparser.js`.
 2. `git filter-branch` should solve it, but we should make sure that no parser-related commits get removed.

/cc @Plaristote @akreuzkamp, please weigh in.

Plaristote:
The first proposition, of having qmlweb-parser being an optional runtime dependency for qmlweb, is the most important thing to me in this.

There's a first problem to solve about that : I think the parser is, right now, a mandatory dependency for qmlweb, even if we split it away. Because the parser files, right now, not only parser qml and javascript, to produce a tree... it also implements the code walking the tree.

So, if I'm not mistaken, the first step we have to do is to move the walker away from the parser. Again, if I'm not mistaken, this should remove the requirement of QMLBinding within qmlweb-parser (because, as long as QMLBinding is in the parser, it will be a mandatory dependency for qmlweb).

akreuzkamp:
I request to wait with splitting off the parser until we have decided about the long term future of QmlWeb. I think, we've got some things to discuss about the direction we want to go, like going all precompiled (if so, how) or just keep things as they are. I think, that discussion has some impact on this one.

akreuzkamp:
still splitting the convertToEngine() part off the parser seems like a good idea. (That part is only Lauris and a bit of my work, so we can easily change the license)

labsin:
Splitting it would also be useful to implement web workers in the future.

still splitting the convertToEngine() part off the parser seems like a good
idea. (That part is only Lauris and a bit of my work, so we can easily
change the license)

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly or view it on GitHub
<https://github.com/qmlweb/qmlweb/issues/173#issuecomment-209013014>

[...]

PR #215, opened 2016-04-27
parser: Move QML runtime elements to qmlstructure.js

ChALkeR:
This is the first step towards #173.

Basicaly, `parser.js` now exports only `qmlweb_parse`, everything else is internal.

Elements from `qmlstructure.js` and `QMLBinding` would be needed at runtime, `qmlweb_parse` (and everything else from `parser.js`) won't be, if the user would use pre-parsed files.

In future, everything but `parser.js` would be kept in the main repo, `parser.js` would be made a stand-alone library and an optional requirement (auto-loaded, though).

/cc @Plaristote, @akreuzkamp, @lpaimen.

@akreuzkamp, @lpaimen — could you confirm that I got the copyrights correct? There seems to be no code from UglifyJS in the new `qmlstructure.js`, please correct me if I'm wrong.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,6 +14,7 @@ var karma = require('karma');
 var qtcoreSources = [
   'src/qtcore/qml/QMLBinding.js',
   'src/qtcore/qml/lib/parser.js',
+  'src/qtcore/qml/lib/qmlstructure.js',
   'src/qtcore/qml/lib/import.js',
   'src/qtcore/*.js',
   'src/qtcore/qml/qml.js',
--- a/src/qtcore/qml/lib/parser.js
+++ b/src/qtcore/qml/lib/parser.js
@@ -53,13 +53,8 @@
  *
  * Exports:
  *
- * - QMLBinding(src, tree) to pass qml bindings along.
- *
  * - parseQML(src) -- parses QML source and returns it as output tree expected
  *   by the QML engine
- *
- * - qmlweb_parse(src) -- parses QML or JS source and returns tree a la uglifyjs parser.
- *   Currently used for debugging purposes.
  */
 
 // Object cloning for debug prints.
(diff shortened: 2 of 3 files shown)


ChALkeR:
This doesn't work with `nowParsingFile`, will fix that.

ChALkeR:
Update: `nowParsingFile` is now a property of `qmlweb_parse()`.
We should make that a parameter in the future, but I don't want to touch that in this PR.

Plaristote:
LGTM

ChALkeR:
This does not conflict with merges, btw.

ChALkeR:
Another way to merge this would be to get a confirmation from @mishoo that there is no his code in the new `qmlstructure.js` file [here](https://github.com/qmlweb/qmlweb/pull/215/files#diff-8a1e2756d324ad38ababb558203ab8f7).

@mishoo: a short history: the qml parser (`parser.js`) here was based on your UglifyJS, but in the first public commit @lpaimen included the new qml-related code directly inside the `parser.js` file. We are now refactoring thigs and want to split those two (the basic parser in and the qml-related walker that converts the parsed tree to the format understood by the engine in) into separate files (`parser.js` and `qmlstructure.js` accordingly) and perhaps even separate repos later. The question is whether did I get the copyrights in the new `qmlstructure.js` file correct (link at the top of this comment), i.e. does it include any code that you authored or not?

ChALkeR:
@lpaimen, @mishoo, I tried hard to trace the origin of the code that is being moved to `qmlstructure.js`, and that looks like that specific code wasn't inherited from UglifyJS.

Most of that is qml-specific, and I wasn't able to find any traces of those methods in UglifyJS sources.

Parser commits rebased on UglifyJS are present at https://github.com/qmlweb/qmlweb-parser/commits/master, and it could be seen which parts were inherited from UglifyJS.

I am going to merge this PR, please leave a comment here if you will have any considerations or updates about this.

PR #216, opened 2016-04-28
Autoload parser

ChALkeR:
This is the second step towards #173 and is based on #215.

This auto-loads `qmlweb.parser.js` when it's required, and changes some build steps so that the whole `qt.js` file is now processed with `babel` at once, not per-component.


ChALkeR:
Rebased.

src/qtcore/qml/lib/qmlstructure.js
```diff
@@ -238,7 +238,27 @@ function convertToEngine(tree) {
...
+
+  console.log('Loading parser...');
+  var tags = document.getElementsByTagName('script');
+  for (let i in tags) {
+    if (tags[i].src && tags[i].src.indexOf('/qt.') !== -1) {
+      let src = tags[i].src.replace('/qt.', '/qmlweb.parser.');
+      // TODO: rewrite to async loading
+      let xhr = new XMLHttpRequest();
+      xhr.open('GET', src, false);
+      xhr.send(null);
```
ChALkeR: Perhaps should check for errors after this line.

PR #218, opened 2016-04-29
@igosoft merge: QMLEngine: return root component from load{File,QML,QMLTree}

ChALkeR:
It seems that nothing does use this in @igosoft fork, so I am not sure if we need this.

~~Based on #193, review only the last commit here.~~ *Update: rebased on master instead*.

This is a cherry-pick of a6618d4bb56ffb97a75d12c02fc5cdea91e2bda6 with `loadQML` support added for consistency.

/cc @igosoft @Plaristote @akreuzkamp @pavelvasev

Refs: #128.

### DIFF
--- a/src/qtcore/qml/QMLEngine.js
+++ b/src/qtcore/qml/QMLEngine.js
@@ -97,12 +97,12 @@ QMLEngine = function (element, options) {
         this.$basePath = this.extractBasePath(file);
         this.ensureFileIsLoadedInQrc(file);
         tree = convertToEngine(qrc[file]);
-        this.loadQMLTree(tree, parentComponent, file);
+        return this.loadQMLTree(tree, parentComponent, file);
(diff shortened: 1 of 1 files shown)


ChALkeR:
Rebased on master instead of #193, this is not blocked anymore.

Plaristote:
We don't need this, but I'm not hostile to having it merged. It could help writing tests for these functions in the future (but then, maybe we should wait for those tests to be written).

ChALkeR:
@Plaristote Thanks! I'm not convinced that we need it yet, I think will not merge this until someone who is in favor of this change comes up.

ChALkeR:
On a second thought, I am going to merge this now, as there are no objections and @igosoft needed that code for some reason.

We can cleanup things after the merges, so let's not hold this further.

PR #220, opened 2016-04-30
@igosoft merge: Do not return values from property setters

ChALkeR:
There is no need to return things from setters, those are not used — `a = b` always returns `b` (i.e. the value being assigned), returning anything from the setter does not affect that.

This merges abdc0e0adfd6529c6f6989a1a780845bfb628393, ref: #128.

LGTM, going to merge shortly.


ChALkeR:
Merged.

PR #221, opened 2016-04-30
@igosoft merge: Do not perform layoutChildren for empty positioners

ChALkeR:
This merges 73c6e57, ref: #128.

I am not conviced with this patch. Removing the children should update implicit size, and with this PR that would not happen, as it seems to me.

/cc @igosoft

### DIFF
--- a/src/qtcore/qml/elements/QtQuick/Column.js
+++ b/src/qtcore/qml/elements/QtQuick/Column.js
@@ -5,6 +5,7 @@ function QMLColumn(meta) {
 QMLColumn.prototype.layoutChildren = function() {
     var curPos = 0,
         maxWidth = 0;
+    if (this.children.length === 0) return;
     for (var i = 0; i < this.children.length; i++) {
         var child = this.children[i];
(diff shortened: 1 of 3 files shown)


ChALkeR:
I am not conviced with this patch. Removing the children should update implicit size, and with this PR that would not happen, as it seems to me.

igosoft:
As far as I remember removing one child update size so removing all children must update as well.

ChALkeR:
@igosoft When all children (or the last child) is removed with this patch, layoutChildren would do nothing and would not update implcit size.

Am I correct? If yes, should we reject this, then?

igosoft:
Let me check behaviour in Qt but probably you are right. I'll do it within 2-3 days.

ChALkeR:
@igosoft Thanks!

igosoft:
@ChALkeR Please reject my changes. You are right, implicit size is not properly update.

ChALkeR:
@igosoft Thanks!

PR #222, opened 2016-04-30
tests: Add some more missing initialize tests

ChALkeR:
Also this adds an empty `'QtMobility 1.2'` initialize tests category, to remind us that we don't have any tests there.

/cc @henrikrudstrom, @Plaristote


Plaristote:
LGTM
(and next test will be the first hundredth !)

Issue #223, opened 2016-04-30
Merge remaining @igosoft patches


ChALkeR:
These are the remaining @igosoft patches.

This is a complete list of individual patches that sum up to the original patch by @igosoft.
It replaces both #128 and #62.

Note that the order might be not correct, and some patches that are closer to the bottom could require some patched that are closer to the top (and other way around).

 - [x] c819a71 Rebuild ./lib — **Rejected**.
 - [x] 71eb280 src/uglify: Make three constants local — **Rejected**.
 - [x] 566df0c Replace eval with new Function in import.js — **Rejected**.
 - [x] a483a33 parser.js: Add "string" and "num" — **Rejected**.
 - [x] aa0a3f4 Change Item overflow style from "hidden" to "auto" — **Rejected**.
 - [x] 24c3ae4 Add some unused commented out code — **Rejected**.
 - [x] 41e7621 unset item.parent inside $delete $tidyupList — **Rejected**.
 - [x] 85d3f04 Code style changes and microoptimization — **Rejected**.
 - [x] fb6d78b Unwanted Loader changes (mostly style) — **Rejected**.
 - [x] 6b018eb Unwanted changes and microptimizations — **Rejected**.
 - [x] 15265a9 Some code structuring in ListModel — **Rejected**.
 - [x] 6c9d44a QMLPositioner: bind to implicit size changes — **Rejected** in #219.
 - [x] f507944 Repeater updates — **Merged** in #248.
 - [x] 691330f ListModel updates — **Merged** in #242.
 - [x] 46c6f2a Finalize QMLComponent imports in QML Object constructor — **Postponed** in #246.
 - [x] 2308bcc move default property updates to the top in applyProperties — **Merged** in #225.
 - [x] 31afc1f Item parent fixes — **Postponed** in #245.
 - [x] af7f263 Remove this.dom.qml property — **Merged** in #224.
 - [x] cc4a8db Positioners implicit size fixes — **Discarded** by @igosoft, comments inside.
 - [x] 6b7593a Update Loader — **Merged** in #247.
 - [x] d6dfbeb Context fixes — **Postponed** in #244. Some changes already addressed by c13919e.
 - [x] f46c60c Property binding fixes — **Discarded** by @igosoft, comments inside.
 - [x] f6c9c6d Text: unify implicit size updates, minor fixes — **Merged** in fbe7c26d68.
 - [x] 50c9d8e Do not perform layoutChildren for empty positioners — **Rejected** in #221.
[...]

PR #224, opened 2016-04-30
@igosoft merge: Remove this.dom.qml property

ChALkeR:
This merges commit af7f263, ref: #223.

Not sure if we want this, but it's not used anywhere expect for being checked in the tests (that check is also removed here).

/cc @igosoft, @akreuzkamp, @Plaristote.


Plaristote:
It has had it's uses, but I don't think we need it anymore.
LGTM.

PR #225, opened 2016-04-30
@igosoft merge: move default property updates to the top in applyProperties

ChALkeR:
This merges commit 2308bcc, ref: #223.

We might need some more explanations on this, but this is likely related to loader per @igosoft comment.

/cc @igosoft, @akreuzkamp, @Plaristote.


ChALkeR:
No objections have been rised. Rubber-stamp LGTM.

PR #226, opened 2016-05-01
jsparser: Replace eval with codegen

ChALkeR:
This is a follow-up to #214 that removes the remaining `eval()` and replaces it with code generation.

No more evals in code :wink:.

/cc @Plaristote

### DIFF
--- a/src/qtcore/qml/lib/jsparser.js
+++ b/src/qtcore/qml/lib/jsparser.js
@@ -4,10 +4,7 @@
       with ($context) {
         ${jsData.source}
       }
-      for (var i = 0 ; i < jsData.exports.length ; ++i) {
-        var symbolName = jsData.exports[i];
-        $context[symbolName] = eval(symbolName);
-      }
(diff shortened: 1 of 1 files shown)


Plaristote:
Shit. I'm learning stuff. Did not know that syntax existed in javascript.
LGTM !

ChALkeR:
@Plaristote It's new, and the browser support isn't perfect yet (check http://kangax.github.io/compat-table/es6/, this patch uses `template literals` and `arrow functions`).

We have Babel enabled since #183, though, and it transplits
```js
(new Function('jsData', '$context', `
  with ($context) {
    ${jsData.source}
  }
  ${jsData.exports.map(sym => `$context.${sym} = ${sym};`).join('')}
`))(jsData, $context);
```
into
```js
new Function('jsData', '$context',
  '\n      with ($context) {\n        ' +
  jsData.source +
  '\n      }\n      ' +
  jsData.exports.map(function (sym) {
    return '$context.' + sym + ' = ' + sym + ';';
  }).join('') +
  '\n    '
)(jsData, $context);
```

PR #227, opened 2016-05-01
import: Remove unused importJs and readExports methods

ChALkeR:
`importJs()` is unused and is very similar to `importJavascriptInContext`.
`readExports()` was used only by `importJs`, so it's actually unused. It's also similar to `jsparse()`.

Note that `importJs` actually implemented getters-setters, which we don't have in our `importJavascriptInContext()`.

/cc @akreuzkamp, @Plaristote, @lpaimen, @pavelvasev.


Plaristote:
Well, I like our new method better. Less duplication of code.
About the usefulness of getter/setters: javascript variables should behave like properties. If they change, then any property linked to it should be updated too, right ?
That's not how it works right now (obviously since there's no getter/setters). I can't check the original QML behavior right now (seems like my setup is broken), but I'm assuming it should work this way. If someone can concur, we should make an issue out of this.

LGTM otherwise.

ChALkeR:
@Plaristote I opened https://github.com/qmlweb/qmlweb/issues/228 for that.

Issue #230, opened 2016-05-03
Notice: contributors activity and PR review process


ChALkeR:
/cc @qmlweb/collaborators, @qmlweb/core.

Given the current situation with PR review, I am going to land less important and less controversial changes directly on `master` for the time being.

I will still open PRs for merges of other peoples' code and for changes that are controversial, or where I would want to double-check that things are ok.

This issue will be open until the activity will rise again =).

ChALkeR:
Closing this notice and posponing the issue until we will have more active contributors.

machinekoder:
@ChALkeR It is still useful to create PRs, even if you merge them without further public reviewing. So the procedure stays the same when there are more active contributors and you get useful CI results before merging into the master branch.

PR #231, opened 2016-05-03
tests: per-file coverage

ChALkeR:
Instrumentation is now done on the gulp side.
This makes the coverage look nice =).


ChALkeR:
Per #230, merging without review.

ChALkeR:
This was merged, the status is not correct here.

PR #233, opened 2016-05-06
@pavelvasev merge: Merge recursive props

henrikrudstrom:
merged recursive properties from @pavelvasev branch
@ChALkeR

### DIFF
--- a/src/qtcore/qml/QMLEngine.js
+++ b/src/qtcore/qml/QMLEngine.js
@@ -477,8 +477,9 @@ QMLEngine = function (element, options) {
 
     this.$initializePropertyBindings = function() {
         // Initialize property bindings
-        for (var i = 0; i < this.bindedProperties.length; i++) {
-            var property = this.bindedProperties[i];
+        // we use `while`, because $initializePropertyBindings may be called recursive (because of Loader and/or createQmlObject )
+        while (this.bindedProperties.length > 0) {
+            var property = this.bindedProperties.shift();
             if (!property.binding)
               continue; // Probably, the binding was overwritten by an explicit value. Ignore.
             if (property.needsUpdate)
@@ -494,15 +495,14 @@ QMLEngine = function (element, options) {
                 updateVGeometry.apply( property.obj,[property.val, property.val, property.name] );
             }
         }
-        this.bindedProperties = [];
 
         this.$initializeAliasSignals();
     }
 
     this.$initializeAliasSignals = function() {
         // Perform pending operations. Now we use it only to init alias's "changed" handlers, that's why we have such strange function name.
-        for (var i = 0; i < this.pendingOperations.length; i++) {
-            var op = this.pendingOperations[i];
+        while (this.pendingOperations.length > 0) {
+            var op = this.pendingOperations.shift();
             op[0]( op[1], op[2], op[3] );
         }
         this.pendingOperations = [];
--- a/src/qtcore/qml/QMLEngine.js
+++ b/src/qtcore/qml/QMLEngine.js
@@ -480,8 +480,8 @@ QMLEngine = function (element, options) {
             var property = this.bindedProperties[i];
             if (!property.binding)
               continue; // Probably, the binding was overwritten by an explicit value. Ignore.
-            property.binding.compile();
-            property.update();
+            if (property.needsUpdate)
+                property.update();
         }
         this.bindedProperties = [];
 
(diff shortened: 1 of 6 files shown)


src/qtcore/qml/QMLEngine.js
```diff
@@ -478,8 +478,8 @@ QMLEngine = function (element, options) {
             var property = this.bindedProperties[i];
             if (!property.binding)
               continue; // Probably, the binding was overwritten by an explicit value. Ignore.
-            property.binding.compile();
-            property.update();
+            if (property.needsUpdate)
++               property.update();
```
ChALkeR: Stray `+`.

ChALkeR:
@henrikrudstrom Thanks! Will take a look at this soon.
I will probably just force-push this branch to fix minor nits, ok?

henrikrudstrom:
Thats totally fine!
Fri 6 May 2016 kl. 17:00 skrev Сковорода Никита Андреевич <
notifications@github.com>:

ChALkeR:
https://github.com/qmlweb/qmlweb/commit/595a29fbf1ebca1b2aa9cc8d6bd7d61a4cd78166 got missed, I will add that on top.

src/qtcore/qml/QMLEngine.js
ChALkeR: I better not forget removing this =).

ChALkeR:
I am going to rename `evaluatingPropertyTop` back to to `evaluatingProperty`, that was unneeded change.

ChALkeR:
Btw, this merges 0aab767, 09b8243, 0e80a6e, 595a29f.

src/qtcore/qml/qml.js
```diff
@@ -306,6 +308,11 @@ function applyProperties(metaObject, item, objectScope, componentScope) {
     _executionContext = componentScope;
     for (i in metaObject) {
         var value = metaObject[i];
+        if (i == "id" || i == "$class") { // keep them
+          item[i] = value;
+          continue;
+        }
```
ChALkeR: @pavelvasev, what's going on here? This contradicts the check below, `i == 'id'` below never fires.

src/qtcore/qml/qml.js
```diff
@@ -4,6 +4,9 @@ var GETTER = "__defineGetter__",
     // which property called the getter of a certain other property for
     // evaluation and is thus dependant on it.
     evaluatingProperty = undefined,
+    evaluatingPropertyStack = [],
+    evaluatingPropertyPaused = false,
```
ChALkeR: Unused. Do we need this? /cc @pavelvasev

src/qtcore/qml/QMLEngine.js
```diff
@@ -474,23 +475,33 @@ QMLEngine = function (element, options) {
...
+        // we use `while`, because $initializePropertyBindings may be called recursive (because of Loader and/or createQmlObject )
+        while (this.bindedProperties.length > 0) {
+            var property = this.bindedProperties.shift();
             if (!property.binding)
               continue; // Probably, the binding was overwritten by an explicit value. Ignore.
-            property.binding.compile();
-            property.update();
+            if (property.needsUpdate)
+                property.update();
+            else if (["width","height","fill","x","y","left","right","top","bottom"].indexOf(property.name) >= 0) {
```
ChALkeR: Why are those special? What about anchors? /cc @pavelvasev

ChALkeR:
Ah, and tracking: #32.

ChALkeR:
Rebased.

src/qtcore/qml/qml.js
```diff
@@ -2,6 +2,9 @@
 // which property called the getter of a certain other property for
 // evaluation and is thus dependant on it.
 var evaluatingProperty;
+var evaluatingPropertyStack = [];
+var evaluatingPropertyPaused = false;
```
ChALkeR: This is unused.

ChALkeR:
Ok, rebase didn't work, reverting rebase =).

ChALkeR:
Merging #248 broke this.

ChALkeR:
Ok, I fixed the rebase.

The last two commits LGTM, the first three seem good, I will just merge them for now — we can fix things later after the merges are complete =).

stephenmdangelo:
@ChALkeR So, I've run into a use-case where this breaks geometry updating, as the recursion checks within `update*Geometry` cause the methods to return before properly applying the appropriate updates.  Reverting these commits fixes it.  I haven't had a chance to pin-point it directly, yet.

I'm guessing it's related to the forced `update*Geometry` calls at the end of `$initializePropertyBindings`.  It seems to only call them if they are internal connections between those geometry-related property names and the geometry of the item with those properties, which won't take into account places where anchors or geometry-related properties refer to properties on other items.

PR #237, opened 2016-05-12
Fix Behavior

stephenmdangelo:
This was broken when translation from "!fromAnimation" to using
QMLPropery.ReasonAnimation in 8c5d600d109be4c2bf7373e8e6e5f78d9b996b1d.

Without this change, attaching a Behavior to a property will result in the binding being removed.

### DIFF
--- a/src/qtcore/qml/QMLProperty.js
+++ b/src/qtcore/qml/QMLProperty.js
@@ -87,7 +87,7 @@ QMLProperty.prototype.set = function(newVal, reason, objectScope, componentScope
             return;
         }
     } else {
-        if (!reason != QMLProperty.ReasonAnimation)
+        if (reason != QMLProperty.ReasonAnimation)
             this.binding = null;
         if (newVal instanceof Array)
             newVal = newVal.slice(); // Copies the array


Plaristote:
Thanks for pointing it out. It looks good to me, but the commit title is a bit vague. I'm not exactly sure what that condition used to do though, so I can't really suggest anything else.

Also not entirely sure what the procedure is to pull in requests from other repositories.
Should we pull stephen's branch in a local repository, amend the commit, push it to a new branch, and then make a new pull request ? It would work, but I'm wondering if we really need to make another pull request for the same commit. Any idea @ChALkeR ?

stephenmdangelo:
Yeah, "fix behavior" is really just the side effect of the change. 

Maybe:

"Fix setting a property from an animation removing existing binding"

ChALkeR:
@Plaristote Just cherry-picking and closing this PR manually will do, I suppose.
LGTM, landing.

@stephenmdangelo, thanks!

ChALkeR:
Landed in 70047d8e8074e9e311ca2e2537bd87cf30cbf895.

PR #238, opened 2016-05-15
Use Object.defineProperty on all browsers

ChALkeR:
It's 2016 on my calendar :wink:.
This removes some fallback code from 2012 by @lpaimen.

`Object.defineProperty` is supported by all ES5 browsers, see http://caniuse.com/es5 for those.
Also see https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Browser_compatibility.


ChALkeR:
No objections after 20 days, merging =).

PR #239, opened 2016-05-18
Per import context constructors

stephenmdangelo:
I have added tests that currently fail with master, but work with these changes.

There are still some uses (like in QMLProperty.set) that should be using the perImportContextConstructors, but it is checking for things like "QMLComponent" that are likely going to be imported into every file with "import QtQuick", so it's not really an issue.

### DIFF
--- a/src/engine/QMLEngine.js
+++ b/src/engine/QMLEngine.js
@@ -434,6 +434,7 @@ class QMLEngine {
     const file = `${this.$basePath}${name}.qml`;
     this.ensureFileIsLoadedInQrc(file);
     const tree = QmlWeb.convertToEngine(QmlWeb.qrc[file]);
+    tree.$file = file;
     this.components[name] = tree;
     return tree;
   }
--- a/src/engine/QMLEngine.js
+++ b/src/engine/QMLEngine.js
@@ -171,7 +171,7 @@ class QMLEngine {
       parent: parentComponent
     });
 
-    this.loadImports(tree.$imports);
+    this.loadImports(tree.$imports, undefined, component.importContextId);
     component.$basePath = this.$basePath;
     component.$imports = tree.$imports; // for later use
     component.$file = file; // just for debugging
@@ -199,6 +199,68 @@ class QMLEngine {
       this.userAddedImportPaths = [];
     }
     this.userAddedImportPaths.push(dirpath);
+  }
+
+  /* Add this dirpath to be checked for components. This is the result of
+   * something like:
+   *
+   * import "SomeDir/AnotherDirectory"
+   *
+   * The importContextId ensures it is only accessible from the file in which
+   * it was imported. */
+  addComponentImportPath(importContextId, dirpath, qualifier) {
+    if (!this.componentImportPaths) {
+      this.componentImportPaths = {};
+    }
+    if (!this.componentImportPaths[importContextId]) {
+      this.componentImportPaths[importContextId] = {};
+    }
+
+    const contextComponentImportPaths
+      = this.componentImportPaths[importContextId];
+
+    if (qualifier) {
+      if (!contextComponentImportPaths.qualified) {
+        contextComponentImportPaths.qualified = {};
+      }
+
+      contextComponentImportPaths.qualified[qualifier] = dirpath;
+    } else {
+      if (!contextComponentImportPaths.unqualified) {
+        contextComponentImportPaths.unqualified = [];
+      }
+
+      contextComponentImportPaths.unqualified.push(dirpath);
+    }
+  }
+
+  importSearchPaths(importContextId) {
+    if (!this.componentImportPaths) {
+      return [];
+    }
+    const contextComponentImportPaths
+      = this.componentImportPaths[importContextId];
+    if (!contextComponentImportPaths) {
+      return [];
+    }
+    return contextComponentImportPaths.unqualified || [];
+  }
+
+  qualifiedImportPath(importContextId, qualifier) {
+    if (!this.componentImportPaths) {
+      return "";
+    }
+    const contextComponentImportPaths
+      = this.componentImportPaths[importContextId];
+    if (!contextComponentImportPaths) {
+      return "";
+    }
+    const qualifiedContextComponentImportPaths
+      = contextComponentImportPaths.qualified;
+    if (!qualifiedContextComponentImportPaths) {
+      return "";
+    }
+    return qualifiedContextComponentImportPaths[qualifier] || "";
   }
 
   setImportPathList(arrayOfDirs) {
@@ -253,7 +315,9 @@ class QMLEngine {
     QmlWeb.setupGetterSetter(obj, propName, getter, setter);
   }
 
-  loadImports(importsArray, currentFileDir = this.$basePath) {
+  loadImports(importsArray, currentFileDir = this.$basePath,
+    importContextId = -1)
+  {
     if (!this.qmldirsContents) {
       this.qmldirsContents = {}; // cache
 
@@ -300,56 +364,55 @@ class QMLEngine {
       // TODO if nameIsDir, we have also to add `name` to importPathList() for
       // current component...
 
+      let content = this.qmldirsContents[name];
       // check if we have already loaded that qmldir file
-      if (this.qmldirsContents[name]) {
-        continue;
-      }
+      if (!content) {
+        if (nameIsQualifiedModuleName && this.userAddedModulePaths &&
+            this.userAddedModulePaths[name]
+        ) {
+          // 1. we have qualified module and user had configured path for that
+          // module with this.addModulePath
+          content = QmlWeb.readQmlDir(this.userAddedModulePaths[name]);
+        } else if (nameIsUrl || nameIsDir) {
+          // 2. direct load
+          // nameIsUrl => url do not need dirs
+          // nameIsDir => already computed full path above
+          content = QmlWeb.readQmlDir(name);
+        } else {
+          // 3. qt-style lookup for qualified module
+          const probableDirs = [currentFileDir].concat(this.importPathList());
+          const diredName = name.replace(/\./g, "/");
 
-      let content = false;
-      if (nameIsQualifiedModuleName && this.userAddedModulePaths &&
-          this.userAddedModulePaths[name]
-      ) {
-        // 1. we have qualified module and user had configured path for that
-        // module with this.addModulePath
-        content = QmlWeb.readQmlDir(this.userAddedModulePaths[name]);
-      } else if (nameIsUrl || nameIsDir) {
-        // 2. direct load
-        // nameIsUrl => url do not need dirs
-        // nameIsDir => already computed full path above
-        content = QmlWeb.readQmlDir(name);
-      } else {
-        // 3. qt-style lookup for qualified module
-        const probableDirs = [currentFileDir].concat(this.importPathList());
-        const diredName = name.replace(/\./g, "/");
-
-        for (let k = 0; k < probableDirs.length; k++) {
-          const file = probableDirs[k] + diredName;
-          content = QmlWeb.readQmlDir(file);
-          if (content) {
-            break;
+          for (let k = 0; k < probableDirs.length; k++) {
+            const file = probableDirs[k] + diredName;
+            content = QmlWeb.readQmlDir(file);
+            if (content) {
+              break;
+            }
           }
         }
+        this.qmldirsContents[name] = content;
       }
 
-      if (!content) {
-        console.log(
-          "qmlengine::loadImports: cannot load qmldir file for import name=",
-          name
-        );
-        // save blank info, meaning that we failed to load import
-        // this prevents repeated lookups
-        this.qmldirsContents[ name ] = {};
-
-        // NEW
-        // add that dir to import path list
-        // that means, lookup qml files in that failed dir by trying to load
-        // them directly
-        // this is not the same behavior as in Qt for "url" schemes,
-        // but it is same as for ordirnal disk files.
-        // So, we do it for experimental purposes.
+      /* If there is no qmldir, add these directories to the list of places to
+       * search for components (within this import scope). "noqmldir" is
+       * inserted into the qmldir cache to avoid future attempts at fetching
+       * the qmldir file, but we always need to the call to
+       * "addComponentImportPath" for these sorts of directories. */
+      if (!content || content === "noqmldir") {
         if (nameIsDir) {
-          this.addImportPath(`${name}/`);
+          if (entry[3]) {
+            /* Use entry[1] directly, as we don't want to include the
+             * basePath, otherwise it gets prepended twice in
+             * createComponent. */
+            this.addComponentImportPath(importContextId,
+              `${entry[1]}/`, entry[3]);
+          } else {
+            this.addComponentImportPath(importContextId, `${name}/`);
+          }
         }
+
+        this.qmldirsContents[name] = "noqmldir";
 
         continue;
       }
--- a/src/engine/qml.js
+++ b/src/engine/qml.js
@@ -149,7 +149,8 @@ function mergeObjects(obj1, obj2) {
   return mergedObject;
 }
 
-const perContextConstructors = {};
+var importContextIds = 0;
+const perImportContextConstructors = {};
 
 function loadImports(self, imports) {
   constructors = mergeObjects(modules.Main, null);
@@ -168,7 +169,8 @@ function loadImports(self, imports) {
     else
       constructors = mergeObjects(constructors, moduleConstructors);
   }
-  perContextConstructors[self.objectId] = constructors;
+  self.importContextId = importContextIds++;
+  perImportContextConstructors[self.importContextId] = constructors;
 }
 
 function inherit(constructor, baseClass) {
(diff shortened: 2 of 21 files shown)


ChALkeR:
@stephenmdangelo Thanks! Sorry for a late answer. We are currently merging the outdated forks that were based on eariler versions, but that work is mostly done. I will need first to check that there are no conflicts there. Will do that today, hopefully.

tests/QMLEngine/imports.js
```diff
@@ -28,5 +28,17 @@ describe('QMLEngine.imports', function() {
     var qml = load("NoQmldir", this.div);
     expect(qml.value).toBe(67);
   });
+  it("imports are local to file, should succeed", function() {
+    var f = function() {
+      load("LocalToFile/Succeed", this.div);
+    };
+    expect(f.bind(this)).not.toThrowError("No constructor found for WebSocket");
```
ChALkeR: Does this need to be so specific? Perhaps just expect it not to throw any error?

stephenmdangelo: Yeah, that works.  I was just thinking to mirror the "should fail" case, which I think we want to keep specific, so that it doesn't hide any other potential problems?

stephenmdangelo:
@ChALkeR not a problem. I've since found a use-case that is not properly using the "importContextId" I would expect, and still need to hunt down the root cause. So, while this does fix some issues, it doesn't seem to be 100% correct yet. You may want to hold off on reviewing until I fix that issue.

I have also added additional support on top of this change for qualified module imports (e.g. import <ModuleIdentifier> <Version.Number> [as <Qualifier>]). Are you good with me pushing those commits on top of this PR, since they are related/dependent?

ChALkeR:
> So, while this does fix some issues, it doesn't seem to be 100% correct yet. You may want to hold off on reviewing until I fix that issue.

Marking as work in progress for now, then =).

ChALkeR:
@stephenmdangelo There are some import and context-related commits in the merges, see the [merge](https://github.com/qmlweb/qmlweb/pulls?q=is%3Apr+is%3Aopen+label%3Amerge) label.

I would prefer to sort those out first.

stephenmdangelo:
@ChALkeR that's cool. I see the flurry of activity. I'll wait for it all to settle and then rebase and sort out the remaining issue. 

I'm glad to see #247 in, as I had manually grabbed that commit and made some changes to it for this stuff to work inside Loaders.

ChALkeR:
@stephenmdangelo I merged most of the forks, except for four commits at https://github.com/qmlweb/qmlweb/labels/merge and the `kde-realinheritance` branch that can't be merged as-is.

Could you rebase this?

stephenmdangelo:
@ChALkeR I actually just rebased today. I'll double check it all tomorrow and push an update. 

One issue with the rebase was that I was using the now-removed context argument being passed to Qt.createComponent when creating the sourceComponent from the source URL within Loader so that these changes work there as well. 

I'm thinking that instead I'll move the implementation of Qt.createComponent to a separate function, with a required context argument. @igosoft, thoughts? (since you did the Loader changes).

I also have a fix for component caching that is required for this to work properly within Loaders and Repeaters:  https://github.com/stephenmdangelo/qmlweb/commit/17503d17718b43e7ba763bc3fb69b18043126f51

I'll ensure that this is in the PR too.

stephenmdangelo:
So I've rebased this and ensured the tests still pass.  I had to revert a change from 2c7b24026bd8961d7ff5d8a843b57242bb8aad86 that caused this to break, as when `createObject` is called within `construct`, it would no longer create a new context, so the importContextId of the newly created Component in `construct` would overwrite those in the Component it is being constructed within.  You can see an example of this in the "Succeed" test I've added here.

I don't know if that change in 2c7b24026bd8961d7ff5d8a843b57242bb8aad86 is required for anything? As you can see, all tests pass without it, and it doesn't REALLY seem to match the description of the other changes in the commit.  Maybe @pavelvasev has some more insight into that?

I'm still going to add a test that uses Loaders, as I believe that that is where I found the issue I originally mentioned.

ChALkeR:
@stephenmdangelo The original commit by @pavelvasev  was 871fbf03c8dcdda039c5556add2bdfb1b2523c41, and 2c7b240 was what we actually merged after a clean-up and resolving things. See also #32 for where that came from.

ChALkeR:
@stephenmdangelo Hm. Perhaps we merged something in the wrong order and some changes slipped through, as that change was apparently reverted in `fork-pavelvasev` branch, see https://github.com/qmlweb/qmlweb/blob/fork-pavelvasev/src/qtcore.js#L430.

That was done in cfb28a087285c68e5b76147fa1fe9440015bfc63, but it's labeled as being only partially merged in #32, though resolved for some reason.

ChALkeR:
Yes, the only thing missing from cfb28a0 is that one-line reversion.
Could you file a separate pull request for that?

stephenmdangelo:
> Upd: no, there might be something else missing.

@ChALkeR, I shall do nothing atm :+1:

ChALkeR:
@stephenmdangelo On a second glance, the whole 2c7b24026bd8961d7ff5d8a843b57242bb8aad86 should be reverted with that change, except for the two lines that call `createComponent`.

The third parameter of `createObject` introduced in that commit isn't used anywhere except for the line that you reverted.

ChALkeR:
@stephenmdangelo 

Could you file a partial revert of 2c7b240 in a separate PR, with better explanations why is that needed?

The diff should probably look like this:
```diff
diff --git a/src/modules/QtQml/Component.js b/src/modules/QtQml/Component.js
index 40a572d..1a4ea9b 100644
--- a/src/modules/QtQml/Component.js
+++ b/src/modules/QtQml/Component.js
@@ -18,18 +18,16 @@ QMLComponent.getAttachedObject = function() { // static
     return this.$Component;
 }
 
-QMLComponent.prototype.createObject = function(parent, properties, componentContext) {
+QMLComponent.prototype.createObject = function(parent, properties) {
     var oldState = engine.operationState;
     engine.operationState = QMLOperationState.Init;
     // change base path to current component base path
     var bp = engine.$basePath; engine.$basePath = this.$basePath ? this.$basePath : engine.$basePath;
 
-    if (!componentContext) componentContext = this.$context;
-
     var item = construct({
         object: this.$metaObject,
         parent: parent,
-        context: componentContext ? Object.create(componentContext) : new QMLContext(),
+        context: this.$context ? Object.create(this.$context) : new QMLContext(),
         isComponentRoot: true
     });
 
diff --git a/src/qtcore/qml.js b/src/qtcore/qml.js
index 7b5b695..affba58 100644
--- a/src/qtcore/qml.js
+++ b/src/qtcore/qml.js
@@ -242,7 +242,7 @@ function construct(meta) {
             component = Qt.createComponent(meta.object.$class + ".qml");
 
         if (component) {
-            var item = component.createObject(meta.parent, {}, meta.context);
+            var item = component.createObject(meta.parent);
 
             if (typeof item.dom != 'undefined')
                 item.dom.className += " " + meta.object.$class + (meta.object.id ? " " + meta.object.id : "");
```

stephenmdangelo:
Partial revert of 2c7b240 in #274.

stephenmdangelo:
@ChALkeR @pavelvasev I've been working with this some more, and I think that the following is the best course of action, but it's a decently large refactor, so I don't want to go down this path without other people's buy-in. I don't this this impacts #289 in it's current state, much. Feel free to @ any other interested parties.

Seems like this may solve the "QNA" at the top of `QMLEngine.loadImports`.

We should move the logic of `QMLEngine.loadImports` into `QmlWeb.loadImports`. Imports are a Component-level concept, not an Engine-level concept.  There's no need for paths imported like `import "SomeDir/AnotherDir"` to be passed to `QMLEngine.addImportPath` in order to be searched for for modules.  We're just currently sharing the `importPathList` as a place to search for modules AND a place to search for a Component to load.  The former usage is global to the `QMLEngine`, the latter should be local to the Component file.

Also, the global `perImportContextConstructors` and `QMLEngine`-global `qualifiedImportPaths` I've added here don't cover the case of keeping `import "SomeDir/AnotherDir"` in the correct scope.

I think that `QmlWeb.loadImports` should attach both the imported constructors (or, to save memory, just store the module information to look up the actual constructors in the global `modules` during `construct`), as well as any imported paths.  These could then be passed along the various contexts in `createObject` and `construct`, as `importContextId` is done with these changes.

Thoughts?

stephenmdangelo:
So I didn't do all of that refactoring, but I did make

stephenmdangelo:
So I didn't do all of that refactoring, but this now keeps unqualified directory imports in the correct context as well.

stephenmdangelo:
@ChALkeR @akreuzkamp I think this is in a good state, I feel like additional cleanups could always land later (e.g. I haven't tested with qmldir files, as I never use them).  IMHO, it solves a pretty big issue with using qmlweb for anyone coming from Qt QML (as seen in #277).

Is there anything  you would like to see changed to get this merged in?

ChALkeR:
Three commits merged in #274.
Could you rebase this, please?

stephenmdangelo:
Done. Haven't tested it after rebasing, but I've rebased :)

tests/QMLEngine/qml/ImportLocalToFile/ModuleFail.qml
```diff
@@ -0,0 +1,8 @@
+import QtQuick 2.0
+
+Item {
+    ModuleFailSomeComponent {
+        WebSocket {}
+    }
+    WebSocket {}
```
ChALkeR: Is the second WebSocket here on a purpose?

akreuzkamp:
I don't know the import code well enough, to give a well-founded review. Thus I have only looked at the code paths I know. There I haven't seen anything to criticize. So, no concerns from my side :)

tests/QMLEngine/qml/ImportLocalToFile/ModuleFail.qml
stephenmdangelo: I believe I put that there because the current behaviour would have made the first WebSocket valid, while the old behaviour (at least what the documentation states) is that all imports in all files are global. So the second WebSocket was to make sure we don't revert back to that... But now that I'm typing it out, the inner one would protect against that anyway... So I'm good to remove it.

ChALkeR:
`npm run coverage` is failing.

ChALkeR:
Argh, it looks like `npm run coverage` also fails on master. Will take a look soon.

ChALkeR:
Found the error, fixed. Please rebase =).

stephenmdangelo:
I've rebased now.

I opted to split the second `WebSocket` test into it's own "should fail" test.  As I feel that it's entirely possible that we could break just one and not the other (e.g. children end up inheriting the imports from our parent Component (case 1) OR we spill the imports from a child Component into the top-level file (case 2) and this will make it easier to see which one broke.

src/engine/qml.js
```diff
@@ -157,7 +157,8 @@ function mergeObjects(obj1, obj2) {
   return mergedObject;
 }
 
-const perContextConstructors = {};
+var importContextIds = 0;
+const perImportContextConstructors = {};
```
stephenmdangelo: @ChALkeR This may be a future cleanup, but thoughts on keying this off of the Component filename instead of an `importContextId`? I feel it could make things easier to debug, as it would be pretty clear if a chunk of code inherited the imports of a different file.

src/engine/QMLEngine.js
```diff
@@ -439,6 +442,52 @@ const QMLEngine = function (element) {
...
     }
 
+    /* Add this dirpath to be checked for components. This is the result of something like:
+     *
+     * import "SomeDir/AnotherDirectory"
+     *
+     * The importContextId ensures it is only accessible from the file in which
+     * it was imported. */
+    this.addComponentImportPath = function ( importContextId, dirpath, qualifier ) {
+        if ( !this.componentImportPaths ) this.componentImportPaths = {};
```
stephenmdangelo: @ChALkeR Should this be `$componentImportPaths`?

stephenmdangelo:
I've rebased this onto master with the QMLEngine refactor. Any objections?

pavelvasev:
Could you please keep old behavior as an option? In practice it seems to be handy to skip imports when they already imported by some other component.

The option might be even not explicit, but in a way of redefinition of some function.

akreuzkamp:
@pavelvasev I'm very much against keeping an option for the old behavior. It's really bad style and really not an issue to add the imports.
To be honest, I'd rather do the opposite... The only way I'd support deviating from the official QML implementation is by *also* forbidding access to *properties* from outer components. It's considered bad style as well. The recommended way to keep a global state is singletons (which aren't yet supported by QmlWeb, though).

Redefining some function is always an option, but obviously at your own risk. But come on, swallow the pill and invest the 30 minutes to fix all your imports ;)

pavelvasev:
Hi Anton! (@akreuzkamp) ! 

I just ask for an option. Why do you refuse me with it?...

stephenmdangelo:
@pavelvasev The problem is that the current behaviour isn't how it is described.  It's described as "all imports are global", but certain imports in files will actually stomp on other imports.  If you want to have a branch that maintains "all imports are global", it should be enough to just remove the `importContextIds++` line, and every context/component will share the same imports.

akreuzkamp:
You know I don't like bad coding style :D (you remember our discussions in Randa about pragmatic solutions vs good style :) ) It wasn't meant to be insulting, though. Sorry for that.

I don't like the thought of having specific code for something that we don't need to support, but need to put 500 warning-, skull- and radioactivity-signs in front of it...

You're of course welcome to redefine the function in your own code :) Thanks @stephenmdangelo for giving the hint, how to do so :) Still my recommendation is not to do so.

ChALkeR:
@akreuzkamp: https://github.com/facebook/react/blob/62f64a84a43b9d8288df5136b816d76dcc7e3321/src/umd/ReactUMDEntry.js#L21-L22 :wink:

stephenmdangelo:
So, no objections? :)

pavelvasev:
Thank you guys! :)

21 июля 2016 г. 23:04 пользователь "stephenmdangelo" <
notifications@github.com> написал:

stephenmdangelo:
@ChALkeR So this is rebased on master again. Any objections to merging it? I'd love to stop rebasing this :).

src/engine/qml.js
ChALkeR: Yes, I'm also thinking about moving this, perhaps even to properties.
`src/engine/qml.js` has some issues, and perhaps it should be even split or renamed.

src/modules/QtQuick/Item.js
```diff
@@ -54,7 +54,12 @@ registerQmlType({
       this.dom.style.position = "absolute";
     }
     this.dom.style.pointerEvents = "none";
-    this.dom.className = `${meta.object.$class}${this.id ? ` ${this.id}` : ""}`;
+    // In case the class is qualified, only use the last part for the css class
+    // name.
+    const classComponents = meta.object.$class.split(".");
+    this.dom.className
+      = `${classComponents[classComponents.length - 1]}${this.id ?
```
ChALkeR: If something like this does not fit into one line, it's better to just introduce a new varable and name it, for this to be more readable. Line length limit is exactly for that — it enforces making the code more readable and less complex.

`const classComponent = meta.object.$class.split(".").pop();` should be fine here.

Will fix this myself while landing.

src/modules/QtQml/Component.js
```diff
@@ -70,10 +70,17 @@ class QMLComponent {
     const bp = engine.$basePath;
     engine.$basePath = this.$basePath ? this.$basePath : engine.$basePath;
 
+    const context = this.$context ? Object.create(this.$context)
+      : new QMLContext();
+
+    if (this.importContextId !== undefined) {
+        context.importContextId = this.importContextId;
```
ChALkeR: Does this pass the linter?

ChALkeR: Ah, it was amended in a follow-up commit.

I will partially squash while landing.

ChALkeR:
Ok, rubber-stamp LGTM, merging.
Thanks and sorry for the delay! =)

ChALkeR:
Ow, it fails the tests again. I will need to take a closer look.

ChALkeR:
The only failing tests are the ones from d897327 — @stephenmdangelo, those are yours.

ChALkeR:
Branch [pr-239-clean](https://github.com/qmlweb/qmlweb/compare/master...pr-239-clean) has a cleaned up versions of these two commits, you could use those.

stephenmdangelo:
@ChALkeR thanks for reviewing. I have an additional commit that fixes Loaders, I'll put that to the clean branch tomorrow. I've also noticed that my changes to "loadImports" cause missing "qmldir" files to get re-requested every time their directory is imported. I'll put that fix in too.

ChALkeR:
@stephenmdangelo Thanks!

stephenmdangelo:
@ChALkeR Should be good now. Please take another look when you have a chance.

I replaced Loader's use of `construct` with `$createObject` (as the tests were failing because `importContextId` was not being set on the context before `construct` was called) and expanded the Loader tests to check that import contexts are correct.

And included the fix for excess checks for non-existant `qmldir`.

src/engine/QMLEngine.js
```diff
@@ -201,6 +201,68 @@ class QMLEngine {
...
+      return [];
+    }
+    return contextComponentImportPaths.unqualified || [];
+  }
+
+  qualifiedImportPath(importContextId, qualifier) {
+    if (!this.componentImportPaths) {
+      return "";
+    }
+    const contextComponentImportPaths
```
ChALkeR: From here, downwards:
```js
const paths = this.componentImportPaths[importContextId];
if (!paths || !paths.qualified) {
  return "";
}
return paths.qualified[qualifier] || "";
```

I will perhaps fix this after landing.

ChALkeR:
Rubber-stamp LGTM. Landing.

ChALkeR:
Thanks!

PR #241, opened 2016-06-06
@pavelvasev merge: propagate properties of component root objects to outer context

ChALkeR:
This merges commit 561b202 by @pavelvasev.

Refs: #32.

/cc @pavelvasev, could you explain what is going on here?

### DIFF
--- a/src/qtcore/qml/qml.js
+++ b/src/qtcore/qml/qml.js
@@ -288,8 +288,11 @@ function createProperty(type, obj, propName, options = {}) {
       }
     }
     setupGetterSetter(obj, propName, getter, setter);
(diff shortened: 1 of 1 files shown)


ChALkeR:
This breaks tests in the current state. Perhaps it depends on other commits or is broken.

ChALkeR:
Ok, this shouldn't block 0.1.0 and cleanups — it's a one-liner, and it breaks the tests. So I'm hoping that @pavelvasev will either fix this, or will discard it =).

ChALkeR:
Reverted in cfb28a087285c68e5b76147fa1fe9440015bfc63, closing.

PR #242, opened 2016-06-08
@igosoft merge: ListModel updates

ChALkeR:
This merges commit 691330f, ref: #223.

This adds support for `append()`-ing an array of elements to `ListModel`, but somewhy doesn't add support for `insert()`-ing an array of elements. Qt QML supports both, afaik.

/cc @igosoft


Plaristote:
LGTM, I added test for this (and the rest of ListModel) in #251

PR #244, opened 2016-06-08
@igosoft merge: Context fixes

ChALkeR:
This merges commit d6dfbeb68afbd32c48c714ba48d3b3be41f36ab7, ref: #223.

One change here was excluded (in `Component.js`), it looks like it was already addressed by c13919e.
The remaining changes are probably also not needed anymore, but I will take a further look on this.

/cc @igosoft, @akreuzkamp

### DIFF
--- a/src/qtcore/qml/QMLEngine.js
+++ b/src/qtcore/qml/QMLEngine.js
@@ -299,7 +299,7 @@ QMLEngine = function (element, options) {
     }
 
     this.rootContext = function() {
-      return this.rootObject.$context;
+      return global.qmlEngine.doc.$context;
     }
 
     this.focusedElement = (function() {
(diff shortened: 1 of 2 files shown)


ChALkeR:
Ok, this shouldn't block 0.1.0 and cleanups — it's a small change, and it breaks the tests. So I'm hoping that @igosoft  will either fix this, or will discard it =).

akreuzkamp:
Ideally we should discard this in favor of fixing QmlEngine in a way that doesn't need a global qmlEngine variable, but allows for multiple instances of QmlEngine. (Ideally we could split it into a dom-independent QmlEngine and a QQuickView). But until we're there, we can fix this PR, by making QmlEngine a proper singleton, i.e. filling global.qmlEngine not from autoloader, but from QmlEngine's constructor (throwing an error, if it's already filled).

ChALkeR:
I don't think that we need this specific change, it doesn't seem to do anything useful as of now.

Supporting multiple running QMLEngines (as @akreuzkamp described) is being discussed in #364.

PR #243, opened 2016-06-08
@igosoft merge: Item parent fixes

ChALkeR:
This merges commit 31afc1f9ba2286adaddb5559b89201b4fcb86bff, ref: #223.

/cc @igosoft

PR #245, opened 2016-06-08
@igosoft merge: Item parent fixes

ChALkeR:
This merges commit 31afc1f, ref: #223.
First attempt was in #243.

Note that this breaks auto-loading.

/cc @igosoft

### DIFF
--- a/src/modules/QtQuick/Item.js
+++ b/src/modules/QtQuick/Item.js
@@ -349,13 +349,22 @@ function QMLItem(meta) {
         if (engine.rootElement == undefined) {
             // Case 1: Qml scene is placed in body tag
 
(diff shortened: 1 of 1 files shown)


ChALkeR:
Ok, this shouldn't block 0.1.0 and cleanups — it's a small change, and it breaks autoloading. So I'm hoping that @igosoft will either fix this, or will discard it =).

ChALkeR:
This is related to #367 and #369.
I will sort this out before 0.3.0 release.

Plaristote:
It's unclear to me what this PR is supposed to fix... but it looks like it doesn't apply anymore. None of the edited code exists anymore.

PR #246, opened 2016-06-08
@igosoft merge: Finalize QMLComponent imports

ChALkeR:
This merges commit 46c6f2a023cd14ac2eb097e9dd90e6595b9cb2a6, ref: #223.

The code was significantly changed in rebasing (after e47ff51 and follow-up commits) , but it's still a one-liner.

/cc @igosoft, @akreuzkamp 

PTAL.

PR #247, opened 2016-06-08
@igosoft merge: Update Loader

ChALkeR:
This merges commit 6b7593a, ref: #223.

Could depend on #246 and #244.

/cc @igosoft


ChALkeR:
Seems working, rubber-stamp LGTM.

PR #249, opened 2016-06-09
@pavelvasev merge: Component context optimization.. maybe we do not need it?

ChALkeR:
> if we create context for every createComponent call, maybe it is too
heavy... So we introduce option to provide context in createObject
method which overrides it's component context.

This merges 871fbf03c8dcdda039c5556add2bdfb1b2523c41, ref: #32.

/cc @pavelvasev @Plaristote

### DIFF
--- a/src/modules/QtQml/Component.js
+++ b/src/modules/QtQml/Component.js
@@ -18,16 +18,18 @@ QMLComponent.getAttachedObject = function() { // static
     return this.$Component;
 }
 
-QMLComponent.prototype.createObject = function(parent, properties) {
+QMLComponent.prototype.createObject = function(parent, properties, componentContext) {
     var oldState = engine.operationState;
     engine.operationState = QMLOperationState.Init;
     // change base path to current component base path
     var bp = engine.$basePath; engine.$basePath = this.$basePath ? this.$basePath : engine.$basePath;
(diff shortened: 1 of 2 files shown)


Plaristote:
Well, it LGTM.
But there's not much left of the original commit, and I'm wondering if the message is still appropriate for the changes that are made in here. The component cache is gone, the only thing added is the ability to overload the context that createObject usually uses, right ?

ChALkeR:
@Plaristote Thanks! I will amed the commit message.

I will also remove the second parameter of `.createComponent()` — it's not used anymore, and Qt QML does not have it (it has other parameters there).

I will do that in the same commit, because this commit removes the usage of that parameter and I also dropped handling of that parameter when I cherry-pickinged the original commit.

ChALkeR:
Argh. Newly landed `Loader` uses the second parameter of `createComponent`, so I perhaps removed handling of that parameter too early.

ChALkeR:
Ok, everything is good here, landing.

ChALkeR:
For reference, this removes usage of non-standart the second parameter of `Qt.createComponent`, which was removed in ff4002cd7f99a760d47fd6e2d3bc41c33386e9f6.

PR #250, opened 2016-06-10
Scale: Call updateOrigin upon creation to handle 0,0 origin

stephenmdangelo:
Is it preferable to use a local `var self = this;` and reference `self` in `updateOrigin`, or to use `updateOrigin.bind(this)()` as I've done here?


stephenmdangelo:
Option 2 done.

ChALkeR:
Merged, thanks!

PR #252, opened 2016-06-10
Timer set running explicitly

stephenmdangelo:
QML allows `Timer.running` to be set to start/stop a timer: http://doc.qt.io/qt-5/qml-qtqml-timer.html#running-prop

Currently, setting `running` to false works, but not setting `running` to true.

### DIFF
--- a/src/modules/QtQml/Timer.js
+++ b/src/modules/QtQml/Timer.js
@@ -27,19 +27,22 @@ registerQmlType({
         }
     }
 
-    this.start = function() {
-        if (!this.running) {
-            this.running = true;
-            prevTrigger = (new Date).getTime();
+    /* This ensures that if the user toggles the "running" property manually,
+     * the timer will trigger. */
+    this.runningChanged.connect(this, function() {
+        if (this.running) {
+            prevTrigger = new Date().getTime();
             if (this.triggeredOnStart) {
                 trigger();
             }
         }
+    })
+
+    this.start = function() {
+        this.running = true;
     }
     this.stop = function() {
-        if (this.running) {
-            this.running = false;
-        }
+        this.running = false;
     }
     this.restart = function() {
         this.stop();
--- a/tests/QtQuick/Timer.js
+++ b/tests/QtQuick/Timer.js
@@ -28,4 +28,13 @@ describe('QtQuick.Timer', function() {
     var now = new Date();
     qml.start();
   });
+
+  it("can set Timer.running = true to start", function(done) {
+    var qml = load("Running", this.div);
+    qml.yield = function(succeed) {
+      expect(succeed).toBe(true);
+      done();
+    };
+    qml.start();
+  });
 });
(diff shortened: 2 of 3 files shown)


src/modules/QtQml/Timer.js
```diff
@@ -27,14 +27,21 @@ registerQmlType({
...
+        if (self.running) {
+            prevTrigger = new Date().getTime();
             if (this.triggeredOnStart) {
                 trigger();
             }
         }
+    })
+
+    this.start = function() {
+        if (!this.running) {
```
ChALkeR: Is this check needed?

src/modules/QtQml/Timer.js
```diff
@@ -27,14 +27,21 @@ registerQmlType({
...
-    this.start = function() {
-        if (!this.running) {
-            this.running = true;
-            prevTrigger = (new Date).getTime();
+    /* This ensures that if the user toggles the "running" property manually,
+     * the timer will trigger. */
+    this.runningChanged.connect(function() {
+        if (self.running) {
+            prevTrigger = new Date().getTime();
             if (this.triggeredOnStart) {
```
ChALkeR: Mixed `self`/`this`. This should be either an arrow function, i.e. `.connect(() => {`, or use the `.connect(this, function() {` syntax for binding the callback to `this`.

src/modules/QtQml/Timer.js
stephenmdangelo: I'm going to go with no... as `runningChanged` won't get fired, so it'll be a no-op.  It was just me changing as little as possible.  There's also a similar no-op check in `this.stop`.

These are both hold-overs from the very first commit in the repo, so as long as you don't want to keep them for sentimental reasons, they can go.

src/modules/QtQml/Timer.js
stephenmdangelo: You are right, I shall fix this. I'll also add a test for "triggeredOnStart", as it would have found this :)

ChALkeR: We need refactoring, code cleanup, and enforcing lint checks. That would make spotting such things a lot easier =).

ChALkeR: Refactoring and clode cleanup is blocked at merges atm.

stephenmdangelo: Should be good now. I just used the `.connect(this, function() {` method, as it is seems to be more common in the current implementation.

src/modules/QtQml/Timer.js
```diff
@@ -27,19 +27,22 @@ registerQmlType({
...
+    /* This ensures that if the user toggles the "running" property manually,
+     * the timer will trigger. */
+    this.runningChanged.connect(this, function() {
+        if (this.running) {
+            prevTrigger = new Date().getTime();
             if (this.triggeredOnStart) {
                 trigger();
             }
         }
+    })
```
ChALkeR: Missing semicolon, but it's ok for now — style fixed will be done later.

ChALkeR:
LGTM

ChALkeR:
I'm squashing this into a single commit. It's fine to include testcases in the same commits as the real changes, those were done separately only when we were merging old commits that originally came without tests.

ChALkeR:
Merged, thanks!

PR #254, opened 2016-06-10
QMLProperty: Log thrown exceptions when evaluating a binding

stephenmdangelo:
This is more inline with the C++ QML Engine that logs the error, but
continues working.

### DIFF
--- a/src/qtcore/qml/QMLProperty.js
+++ b/src/qtcore/qml/QMLProperty.js
@@ -26,7 +26,11 @@ QMLProperty.prototype.update = function() {
 
     var oldVal = this.val;
     evaluatingProperty = this;
-    this.val = this.binding.eval(this.objectScope, this.componentScope);
+    try {
+      this.val = this.binding.eval(this.objectScope, this.componentScope);
+    } catch (e) {
+      console.log("QMLProperty.update binding error:", e, Function.prototype.toString.call(this.binding.eval))
+    }
     evaluatingProperty = undefined;
 
     if (this.animation) {


ChALkeR:
It does look good to me, but just a question — what's the reason behind logging error as-is, but stringifying the function? Note that browser console pretty-prints functions and errors, afaik. Is there something else going on there?

stephenmdangelo:
Ah yes, I had added the function printing as a way of trying to determine what binding was failing (as some exceptions can look very generic). This was the best way I could get it to show up in Firefox. Obviously even this can only get you so far in finding the source of the exception. 

Ideally, we would mimic QML and print out the file name and line number, but I'm thinking that might be a lot of work to pass it through all of the necessary components. I could look into doing this of you think it's feasible to get line numbers from the parser.

ChALkeR:
Ah, true. Firefox doesn't pretty-print functions in `console.log()`.
For example, compare `console.log('foo', new Error('test'), function(x) { return x * 2 })` output in Firefox and Chromium. Chromium pretty-prints the function header and body, while Firefox prints only `function ()`.

LGTM with converting funtions to strings, then.

> Ideally, we would mimic QML and print out the file name and line number, but I'm thinking that might be a lot of work to pass it through all of the necessary components.

I don't think that we should exactly mimic Qt QML in terms of error reporting to console — for some cases, it's not very nice there (e.g. for some parsing errors). Printing errors/warnings at the same time when Qt prints them is a good thing, but the exact messages could be different.

Adding line numbers to parser could be a good thing, but it will increase the parsed tree size — perhaps we could think of adding a debug flag for that. The current parsed tree structure uses simple lists, though, so adding any additional information there (moreover — optional additional information) would require significant changes in the parser. Perhaps https://github.com/qmlweb/qmlweb-parser/issues/4 is a way to go, but I would postpone that.

src/qtcore/qml/QMLProperty.js
```diff
@@ -26,7 +26,11 @@ QMLProperty.prototype.update = function() {
 
     var oldVal = this.val;
     evaluatingProperty = this;
-    this.val = this.binding.eval(this.objectScope, this.componentScope);
+    try {
+      this.val = this.binding.eval(this.objectScope, this.componentScope);
+    } catch (e) {
+      console.log("QMLProperty.update binding error:", e, Function.prototype.toString.call(this.binding.eval))
```
ChALkeR: Semicolon is missing here, but we will cleanup things later.
Lint checks are not enforced yet in `src/` directory.

ChALkeR:
Thanks!

ChALkeR:
Note: it looks like that wasn't the only place where `binding.eval(` is executed.

ChALkeR:
Also take a look at 32d8a182c141365633ee719e4ceeee08fb09e751.

PR #255, opened 2016-06-10
jsparser: Support importing JavaScript files with ".pragma" statements

stephenmdangelo:
.pragma statements are not currently respected, but they at least won't
cause the rest of the file to fail to import.


ChALkeR:
LGTM, thanks!

ChALkeR:
See #377 for the actual `.pragma library` support.

PR #256, opened 2016-06-10
MouseArea: Set pressed to false on release, even if outside of MouseArea

stephenmdangelo:
This properly mimics QML's behaviour

@ChALkeR Any ideas on how to test this? From a quick Google search, it looks like we would have to use jQuery event handlers and .trigger to simulate events from within Jasmine, but maybe there's a better way.

### DIFF
--- a/src/modules/QtQuick/MouseArea.js
+++ b/src/modules/QtQuick/MouseArea.js
@@ -36,6 +36,11 @@ registerQmlType({
 
     this.dom.addEventListener("click", e => this.$handleClick(e));
     this.dom.addEventListener("contextmenu", e => this.$handleClick(e));
+    const handleMouseUp = () => {
+      this.pressed = false;
+      this.pressedButtons = 0;
+      document.removeEventListener("mouseup", handleMouseUp);
+    };
     this.dom.addEventListener("mousedown", e => {
       if (!this.enabled) return;
       const mouse = this.$eventToMouse(e);
@@ -43,10 +48,7 @@ registerQmlType({
       this.mouseY = mouse.y;
       this.pressed = true;
       this.pressedButtons = mouse.button;
-    });
-    this.dom.addEventListener("mouseup", () => {
-      this.pressed = false;
-      this.pressedButtons = 0;
+      document.addEventListener("mouseup", handleMouseUp);
     });
     this.dom.addEventListener("mouseover", () => {
       this.containsMouse = true;


ChALkeR:
Hah, let's just use PhantomJS API for that.

See http://phantomjs.org/page-automation.html and http://phantomjs.org/api/webpage/method/send-event.html.

stephenmdangelo:
So I've pushed what I'm trying, but I can't seem to get the phantom callback called from within my test.  If you could please let me know if I'm on the right track, or if there's a better approach, that would be awesome.

ChALkeR:
@stephenmdangelo You don't want render tests for that, I suppose.

The coordinates `[0, 0]` are wrong — `this.div` is positioned in a different place. I will add a helper function to get that, based on how our `tests/Render/runner.js` currently calculates the area for the screenshot.

Also, not sure what's up with mouse events, but handing them on `window` object works fine. Perhaps some `karma` issues.

stephenmdangelo:
> You don't want render tests for that, I suppose.

Oh, I know. It's just the only place I was able to get `window.top.callPhantom` within the testing scope (I'm assuming because the Render tests have a custom `runner.js`? I'm new to JavaScript testing stuff.)

> I will add a helper function to get that, based on how our `tests/Render/runner.js` currently calculates the area for the screenshot.

Thanks! That would be appreciated. I'll take a look at the code in `runner.js` and see if I can get something working.

> Also, not sure what's up with mouse events, but handing them on `window` object works fine. Perhaps some karma-related issues.

Sorry, not sure what you mean by this? Are you referring to the actual change, or the testing attempts? Is there a testing method other than using PhantomJS's `sendEvent` I could be using?

ChALkeR:
> Sorry, not sure what you mean by this? Are you referring to the actual change, or the testing attempts? Is there a testing method other than using PhantomJS's sendEvent I could be using?

`sendEvent` works with mouse events, and I can listen to them with `window.on{something}`. I have not made a testcase where this actually works with QML code and I'm not sure if there would be any problems there. Everything is running inside a karma enviroment there, there might be frames, different stuff. Try launching `test-debug` and opening it in the browser window, that could help to figure things out.

stephenmdangelo:
I've rebased this and removed my failed attempts at writing a test case.

ChALkeR:
> I've rebased this and removed my failed attempts at writing a test case.

Could you file them separately, please, and mark those as `work-in-progress`, so that we would have a starting point?

LGTM, merging.

PR #257, opened 2016-06-10
Separate update h v geometry checks

stephenmdangelo:
As you can see, I stumbled upon this one in a super-specific use-case.

I believe I've made the test the simplest possible case, and commented the important bits.


ChALkeR:
LGTM, thanks! Squashing and merging.

PR #261, opened 2016-06-10
Animation: port to the new module definition syntax

ChALkeR:
This ports `Animation` to the new module definition syntax to showcase my proposal in #260.

/cc @Plaristote @akreuzkamp

Issue #260, opened 2016-06-10
Modules definition syntax.


ChALkeR:
Atm, we have mixed modules definition syntax.
 1. [Animation](https://github.com/qmlweb/qmlweb/blob/master/src/modules/QtQuick/Animation.js) uses:
 ```js
  function QMLAnimation(meta) {
      callSuper(this, meta);
      …
  }

  registerQmlType({
    module:   'QtQuick',
    name:     'Animation',
    versions: /.*/,
    baseClass: 'QtQml.QtObject',
    constructor: QMLAnimation
  })
```
 2. [Column](https://github.com/qmlweb/qmlweb/blob/master/src/modules/QtQuick/Column.js) uses:
 ```js
  function QMLColumn(meta) {
      callSuper(this, meta);
  }

  QMLColumn.prototype.layoutChildren = function() {
      …
  }

  registerQmlType({
    module: 'QtQuick',
    name:   'Column',
    versions: /.*/,
    baseClass: 'Positioner',
    constructor: QMLColumn
  });
```
 3. [ListView](https://github.com/qmlweb/qmlweb/blob/master/src/modules/QtQuick/ListView.js) uses:
 
  ```js
  registerQmlType({
    module:   'QtQuick',
    name:     'ListView',
    versions: /.*/,
    baseClass: 'Repeater',
    constructor: function QMLListView(meta) {
      callSuper(this, meta);
      …
    } 
  });
```

We should decide on one format and convert all modules to use that single format.

**Comment by ChALkeR** [comment]
Atm, `registerQmlType` should be on the top — it includes the module name and some important properties (e.g. since 613ca9d0dfd4313c0f6fa248f4538dd09ec81bc7...061230729c9aae23f5c761811a29c8590c7f786e — the properties list).

Prototype functions could be defined after calling `registerQmlType` (that is supported), but I don't see a reason to do that. Also, I don't even see a reason for naming classes.

My proposal here is:
```js
registerQmlType({
  module: 'QtQuick',
  name:   'Column',
  versions: /.*/,
  baseClass: 'Positioner',
  constructor: class {
    constructor(meta) {
      callSuper(this, meta);
      …
    }
    layoutChildren() {
      …
    }
  }
});
```

That should work. Note that it doesn't use `extends`, because we don't know if the base class was already loaded at require time.

ChALkeR:
[...]

PR #263, opened 2016-06-10
Animation: port to the new module definition syntax (stage-2 variant)

ChALkeR:
This ports `Animation` to the new module definition syntax to showcase my proposal in #260.

This is alternative to #261 and uses an even nicer syntax, but requires a ~~`stage-1`~~ `stage-2` transform to support [ES Class Fields & Static Properties](https://github.com/jeffmo/es-class-fields-and-static-properties) ~~`stage-1`~~ `stage-2` proposal. **Upd:** it's `stage-2` now.

`stage-2` means «The committee expects the feature to be developed and eventually included in the standard». See [TC39 Process](https://tc39.github.io/process-document/) for more info on that.

~~I doubt that we want this atm, perhaps we should wait until it reaches `stage-2` or `stage-3`.~~
Not sure if we want to adopt this on the current stage or if we should wait for it to reach `stage-3`, which could take some time.

I landed runtime support for this syntax in 188e9fd (i.e. for passing a `function` to `registerQmlType`).

/cc @akreuzkamp @Plaristote

### DIFF
--- a/src/modules/QtQuick/Animation.js
+++ b/src/modules/QtQuick/Animation.js
@@ -1,19 +1,18 @@
-QmlWeb.registerQmlType({
-  module: "QtQuick",
-  name: "Animation",
-  versions: /.*/,
-  baseClass: "QtQml.QtObject",
-  enums: {
+QmlWeb.registerQmlType(class Animation {
+  static module = "QtQuick";
+  static versions = /.*/;
+  static baseClass = "QtQml.QtObject";
+  static enums = {
     Animation: { Infinite: -1 },
     Easing: QmlWeb.Easing
-  },
-  properties: {
+  };
+  static properties = {
     alwaysRunToEnd: "bool",
     loops: { type: "int", initialValue: 1 },
     paused: "bool",
     running: "bool"
-  }
-}, class {
+  };
+
   constructor(meta) {
     QmlWeb.callSuper(this, meta);
   }
--- a/src/engine/modules.js
+++ b/src/engine/modules.js
@@ -76,7 +76,7 @@ function registerQmlType(options, constructor) {
 
   const descriptor = typeof options === "function" ? {
     module: options.module,
-    name: options.element,
+    name: options.element || options.name,
     versions: options.versions,
     baseClass: options.baseClass,
     enums: options.enums,


ChALkeR:
Status update: [ES Class Fields & Static Properties](https://github.com/jeffmo/es-class-fields-and-static-properties)  is now at `stage-2` aka `Draft` (see https://tc39.github.io/process-document/).

It has now has a formal specification at https://tc39.github.io/proposal-class-public-fields/, and is more likely to be accepted as a future ECMAScript language feature.

We could reconsider this. I updated the PR. @akreuzkamp, @Plaristote, what do you think?

ChALkeR:
Tests broken because of coverage being broken, I will take a look at that if we choose to adopt this syntax.

src/modules/QtQuick/Animation.js
```diff
@@ -1,19 +1,21 @@
-QmlWeb.registerQmlType({
-  module: "QtQuick",
-  name: "Animation",
-  versions: /.*/,
-  baseClass: "QtQml.QtObject",
-  enums: {
+QmlWeb.registerQmlType(class {
+  static module = "QtQuick";
+  static element = "Animation";
```
akreuzkamp: why did you rename that property to element? I think "element" is something you shouldn't use as a name for (almost) anything, because it's too vague... What's wrong about `name`?

akreuzkamp:
does it really work without any changes in `registerQmlType`?

In any way I do like this syntax :)

ChALkeR:
@akreuzkamp The changes in `registerQmlType` landed in 188e9fd, it currently supports all three syntaxes: `options, constructor`, a single constructor with static options, a single options object with a `constructor` property. The last one is old and unused, but it's still there because it's used for internal representation.

src/modules/QtQuick/Animation.js
ChALkeR: `name` is already present for all classes (and functions), it's a string there, and in this specific case it's an empty string because the class is unnamed.

```shell
> (function x(){}).name
'x'
> (class {}).name
''
> (class Foo {}).name
'Foo'
```

Yes, `element` isn't perfect, but I wasn't able to come up with something better. Any suggestions?

ChALkeR: Otoh, we could just write `class Animation {` and keep `.name`. What do you think?

ChALkeR: @akreuzkamp We could do both — use `.name` by default and write `class Animation {` in our code, but allow users to override it with a static property so that they could use anonymous functions there. How should that property be named then?

ChALkeR:
Ok, I'm going to postpone this.

ChALkeR:
Ok, this passes the tests now, and coverage works.

PR #264, opened 2016-06-11
qmlstructure.js, import.js: move copyrights notices to LICENSE     

ChALkeR:
This changes the way how the copyrights notice and the license are included for the qmlstructure.js and import.js files.
    
This does not change the actual copyrights or the license for those files.

/cc @lpaimen @akreuzkamp @pavelvasev — those are your copyrights, so PTAL.

Note: this blocks moving file contents around between files.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -3,8 +3,10 @@ QmlWeb is licensed under the MIT license, as follows:
 """
 MIT License
 
-Copyright (c) 2012 Lauri Paimen
-Copyright (c) 2013 Anton Kreuzkamp
+Copyright (c) 2011, 2012 Lauri Paimen <lauri@paimen.info>
+Copyright (c) 2013 Anton Kreuzkamp <akreuzkamp@web.de>
+Copyright (c) 2015 Pavel Vasev <pavel.vasev@gmail.com> - initial and working
+                                                         import implementation.
 Copyright (c) 2016-2018 QmlWeb contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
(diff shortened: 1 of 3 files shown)


ChALkeR:
This blocks moving file contents around between files.

ChALkeR:
@lpaimen @akreuzkamp @pavelvasev to clarify things: this doesn't alter the actual copyright ownership (it's impossible to do that this way afaik), it also doesn't remove the copyright notices, it just moves the notices to another (global) file.

The reason for doing that is because the code is split in multiple small files now and is concatenated from those files — so those appear in the middle of concatenated code. Also it's harder to maintain copyrights this way (would require year updates in many files, etc).

This needs your approval because I don't think that the license allows to alter how the copyright notice (and license information) is included in such a way without a special permission from the copyright holders (e.g. you).

I will split this into two separate pull requests.

pavelvasev:
I'm ok with that. By the way the main contributors are @akreuzkamp and @lpaimen so what they say..

ChALkeR:
@pavelvasev Thanks! =)

@stephenmdangelo, @Plaristote  it looks like this also needs your review because at least one of those files has your code in it and you are mentioned in the copyrights (`Copyright (c) 2016 QmlWeb contributors`).

stephenmdangelo:
LGTM

pavelvasev:
lgtm

lpaimen:
Very sorry for the big delay! This is fine for me.

akreuzkamp:
Hey, I missed that for almost 6 years... Sorry for that. Yes, it is fine for me :D

akreuzkamp (approved):
lgtm

PR #267, opened 2016-06-13
Connections

stephenmdangelo:
I haven't bothered implementing "ignoreUnknownSignals", as it's just related to how much logging the QML Engine does.

### DIFF
--- a/src/modules/QtQml/Connections.js
+++ b/src/modules/QtQml/Connections.js
@@ -1,0 +1,34 @@
+registerQmlType({
+  module: 'QtQml',
+  name: 'Connections',
+  versions: /.*/,
+  baseClass: 'QtObject',
+  properties: {
+    target: "QtObject",
+    ignoreUnknownSignals: "bool"
+  }
+}, class {
+  constructor(meta) {
+    callSuper(this, meta);
+    this.target = this.$parent
+    this.$connections = {}
+
+    let old_target = this.target;
+    const reconnectTarget = () => {
+      for (var i in this.$connections) {
+        var c = this.$connections[i];
+        if (c._currentConnection && old_target && old_target[i] && typeof old_target[i].disconnect === 'function') {
+          old_target[i].disconnect(c._currentConnection);
+        }
+        c._currentConnection = connectSignal(this.target, i, c.value, c.objectScope, c.componentScope);
+      }
+      old_target = this.target;
+    };
+
+    this.targetChanged.connect(reconnectTarget);
+    this.Component.completed.connect(reconnectTarget);
+  }
+  $setCustomSlot(propName, value, objectScope, componentScope) {
+    this.$connections[propName] = { value, objectScope, componentScope };
+  }
+});
--- a/src/qtcore/qml.js
+++ b/src/qtcore/qml.js
@@ -322,6 +322,30 @@ function setupGetterSetter(obj, propName, getter, setter) {
   });
 }
 
+function connectSignal(item, signalName, value, objectScope, componentScope) {
+    if (!item[signalName]) {
+        console.warn("No signal called " + signalName + " found!");
+        return;
+    }
+    else if (typeof item[signalName].connect != 'function') {
+        console.warn(signalName + " is not a signal!");
+        return;
+    }
+    if (!value.eval) {
+        var params = "";
+        for (var j in item[signalName].parameters) {
+            params += j==0 ? "" : ", ";
+            params += item[signalName].parameters[j].name;
+        }
+        value.src = "(function(" + params + ") { _executionContext = __executionContext;" + value.src + "})";
+        value.isFunction = false;
+        value.compile();
+    }
+    var slot = value.eval(objectScope, componentScope);
+    item[signalName].connect(item, slot);
+    return slot;
+}
+
 /**
  * Apply properties from metaObject to item.
  * @param {Object} metaObject Source of properties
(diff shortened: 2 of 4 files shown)


src/qtcore/qml/qml.js
```diff
@@ -321,6 +321,30 @@ function setupGetterSetter(obj, propName, getter, setter) {
...
+        var params = "";
+        for (var j in item[signalName].parameters) {
+            params += j==0 ? "" : ", ";
+            params += item[signalName].parameters[j].name;
+        }
+        value.src = "(function(" + params + ") { _executionContext = __executionContext;" + value.src + "})";
+        value.isFunction = false;
+        value.compile();
+    }
+    var f = value.eval(objectScope, componentScope);
```
ChALkeR: Could you give the variablie `f` a bit more semantic name here? It's not clear what it is.

src/modules/QtQml/Connections.js
```diff
@@ -0,0 +1,41 @@
+/* Connections */
```
ChALkeR: Not required.

src/modules/QtQml/Connections.js
```diff
@@ -0,0 +1,41 @@
...
+  baseClass: 'QtObject',
+  constructor: function QMLConnections(meta) {
+    callSuper(this, meta);
+
+    createProperty("QtObject", this, "target");
+    createProperty("bool", this, "ignoreUnknownSignals");
+
+    this.target = this.$parent
+
+    var old_target = this.target;
```
ChALkeR: perhaps it should be moved closer to `reconnectTarget`? Also, it should be `let`.

src/modules/QtQml/Connections.js
```diff
@@ -0,0 +1,41 @@
...
+
+    var connections = {}
```
ChALkeR: Could you port this to the new module definition syntax? Including properties, etc.

`connections` should probably be `this.$connections` to make setCustomSlot a function in the prototype.

src/modules/QtQml/Connections.js
```diff
@@ -0,0 +1,41 @@
...
+
+    this.$setCustomSlot = function(propName, value, objectScope, componentScope) {
+      connections[propName] = {
+        "value" : value,
+        "objectScope" : objectScope,
+        "componentScope" : componentScope
+      }
```
ChALkeR: Missing semicolon. Also, `connections[propName] = { value, objectScope, componentScope };`

src/modules/QtQml/Connections.js
```diff
@@ -0,0 +1,41 @@
...
+        if (c._currentConnection && old_target && old_target[i] && typeof old_target[i].disconnect == 'function') {
+          old_target[i].disconnect(c._currentConnection);
+        }
+        c._currentConnection = connectSignal(this.target, i, c.value, c.objectScope, c.componentScope);
+      }
+      old_target = this.target;
+    };
+
+    this.targetChanged.connect(this, reconnectTarget);
+    this.Component.completed.connect(this, reconnectTarget);
```
ChALkeR: second `this` is not required, should be `.connect(reconnectTarget)`.

src/modules/QtQml/Connections.js
```diff
@@ -0,0 +1,41 @@
...
+    this.$setCustomSlot = function(propName, value, objectScope, componentScope) {
+      connections[propName] = {
+        "value" : value,
+        "objectScope" : objectScope,
+        "componentScope" : componentScope
+      }
+    }
+
+    const reconnectTarget = () => {
+      for(var i in connections) {
```
ChALkeR: `for (`

src/modules/QtQml/Connections.js
```diff
@@ -0,0 +1,41 @@
...
+        var c = connections[i];
+        if (c._currentConnection && old_target && old_target[i] && typeof old_target[i].disconnect == 'function') {
```
ChALkeR: `===`

src/modules/QtQml/Connections.js
```diff
@@ -0,0 +1,41 @@
...
+registerQmlType({
+  module: 'QtQml',
+  name: 'Connections',
+  versions: /.*/,
+  baseClass: 'QtObject',
+  constructor: function QMLConnections(meta) {
+    callSuper(this, meta);
+
+    createProperty("QtObject", this, "target");
+    createProperty("bool", this, "ignoreUnknownSignals");
```
ChALkeR: Could you port this to the new module definition syntax? See [modules/QtQuick/Animation.js](https://github.com/qmlweb/qmlweb/blob/master/src/modules/QtQuick/Animation.js).

ChALkeR:
Mostly LGTM, some style nits.

src/modules/QtQml/Connections.js
stephenmdangelo: Huh, who knew?! TIL.

stephenmdangelo:
I believe I've fixed all of those issues now.  Let me know if the port to the new module definition syntax is correct.  I'm guessing there's not any way to have `this.$parent` be the `initialValue` of `target`, as there's not really a `this` to reference at that point.

ChALkeR:
@stephenmdangelo Thanks! Yes, there is no way to default to a dynamic value for now. Perhaps we could add functions support for that (e.g. `target: { type: 'QtObject', initialValue: element => element.$parent }`, but that would block functions being the initial values… Perhaps I need some more time to about that. There are other issues with properties that should be fixed before that, perhaps.

LGTM, merging.

ChALkeR:
Landed in 4aa5ea2...0b40e72.

Thanks!

PR #269, opened 2016-06-14
Split up src/qtcore into src/engine and src/qtbase

akreuzkamp:
This splits up the files in qtcore into qml-engine related stuff and
generic stuff that reimplements QtCore/QtGui functionality.
This implements part of issue #143.
I decided to rename the qtcore directory to qtbase, because it doesn't
only contain QtCore classes, but also QtGui (which in qt5 are both
part of qtbase).

I'm still too stupid to use gulp correctly, so I'm hoping for the CI to test it for me :D


gulpfile.js
```diff
@@ -13,11 +13,12 @@ const karma = require('karma');
...
 const qtcoreSources = [
-  'src/qtcore/QMLBinding.js',
-  'src/qtcore/qmlstructure.js',
-  'src/qtcore/import.js',
-  'src/qtcore/qml.js',
-  'src/qtcore/*.js',
+  'src/enigne/QMLBinding.js',
+  'src/enigne/qmlstructure.js',
+  'src/enigne/import.js',
+  'src/enigne/qml.js',
```
ChALkeR: s/enigne/engine/

ChALkeR:
@akreuzkamp Thanks! This does LGTM, though I am still not sure why, for example, `QMLInteger` is a part of `engine`, but  `QList` (formerly `QMLList`) is a part of `qtbase`. Things like that confused me when I tried to split those myself =).

akreuzkamp:
Ah, dough! I looked at the code of QList again and realized it was actually the wrong decision to put it in qtbase. My motivation was simply that QList has a counterpart in QtCore (as QColor has in QtGui), while QMLInteger hasn't. But what QMLList implements isn't QList, but something that is tightly bound to our qml engine. The qtbase/ folder was supposed to contain stuff, that is useful on its own, but QMLList isn't.

ChALkeR:
@akreuzkamp So, the only adjustment this needs is moving the `qtcore/QList.js` to `engine/QMLList.js`?

LGTM with that change, I suppose.

akreuzkamp:
oh, yes, thanks. I just forgot about it :)

ChALkeR:
@akreuzkamp I rebased this against current master to include the `QmlWeb.js` file.

I also edited the commit message to include the full link to the issue (#{whatever} could break when viewed from forks).

Merging after CI passes.

ChALkeR:
Argh, I made a mistake while rebasing. Will fix in a moment =).

ChALkeR:
Merged, thanks!

Issue #236, opened 2016-05-09
Enums visibility


ChALkeR:
Now, enums are either visible in the global scope (`global.TextInput = {…}`) or at element scope (`this.Text = {…}`).

They should be instead visible on the whole evaluation scope, but not registered on `global`.

henrikrudstrom:
just a thought: Is it an option to use string values instead of enums. then we dont need to pollute the global namespace, makes it easier to debug (no need to look up what the integer value means), and no confusion with the 0-valued enum item and thruthy comparison i.e. `if (!enumValue)`...
(not right now though)

jhihn:
I'm not in QMLWeb, but having used Android SDK,  I caution against using strings. using a proper enum or class is better because then you can add type safety. 
For a more nuanced discussion, see https://stijndewitt.com/2014/01/26/enums-in-javascript/
But I would recommend using var x = EnumType.Value  because that is most source compatible with using objects or proper classes.

ChALkeR:
No, enums shouldn't be strings, they have specific integer values in QML and one could compare enums to integers.

ChALkeR:
cc3cbeacca4871a2177fba967ee6921bfd1ee4f8 adds support for declarative enums definition with the same effect as the most used one for now (i.e. adding the enums inside the component context for that element).

But the good thing is that it allows us to describe enums as 
```js
registerQmlType({
  module: 'Module',
  name: 'Example',
  versions: /.*/,
  baseClass: 'QtQml.Item',
  enums: {
    MyEnum: { Foo: -1, Bar: 0, Test: 10 }
  },
  …
});
```

We could use that later to propertly import enums when the _module_ gets loaded.

PR #271, opened 2016-06-15
New modules syntax — the main part

ChALkeR:
This converts almost everything to (anonymous) classes, `properties:` instead of `createProperty`, and `enums:` instead of manually defining an enum on `global` or `this`.

Spacing changes and other cleanup is out of scope of this PR to make this easier to review.

Note: some property types and initial values are strange here and need a closer look, but that is also out of scope of this PR, I left the actual types and initial values as they were on a purpose.

`Font` requires more work, as it uses a non-standard constructor, so I left it out of this PR.


ChALkeR:
Updated. @akreuzkamp, @Plaristote, PTAL.

Plaristote:
So far so good. Although quick question, why do we need the commit about: `qml: register all enums at global scope for now` ?

I'm assuming it's not useful in QML code, since the enum would already be part of the context, right ? Is this about legacy code still using it in our own javascript ?

ChALkeR:
@Plaristote We don't have proper support for registering enums in the context for now, afaik

Before the new syntax, they were just registered either:
 1. at the `global` object (which is `window` in browsers) — as `Easing` and `TextInput`, this is why `TextInput.Normal` was defined, or
 2. as the property of each object instance — as `Flow`, this is why `Flow { flow: Flow.LeftToRight }` worked — that instance has a `this.Flow = { LeftToRight: 0, TopToBottom: 1 }`.

This branch does both for now — register them as properties of each instance and as properties of the `global` object (if not present already).

The proper way of fixing that would of course involve registering the enums for all imported typed in corresponding contexts, but I wanted to keep that out of scope of this PR for now.

ChALkeR:
@Plaristote See #236 for that, btw.

src/qtcore/QMLProperty.js
```diff
@@ -101,6 +101,7 @@ QMLProperty.prototype.get = function() {
 }
 
 const typeInitialValues = {
+  enum: 0,
```
ChALkeR: This was accidential, will undo it. It's already defined below.

src/modules/QtQuick/BorderImage.js
```diff
@@ -3,31 +3,28 @@ registerQmlType({
...
-        Error: 4
-    }
-
-    createProperty("url", this, "source");
-    createProperty("enum", this, "status", {initialValue: this.BorderImage.Null});
     this.border = new QObject(this);
     createProperty("int", this.border, "left");
     createProperty("int", this.border, "right");
     createProperty("int", this.border, "top");
     createProperty("int", this.border, "bottom");
```
akreuzkamp: Why don't you morph these into the new syntax?

src/modules/QtMultimedia/Video.js
```diff
@@ -1,19 +1,35 @@
...
+  },
+  properties: {
+    source: 'string',
+    duration: 'int',
+    position: 'int',
+    autoPlay: 'bool',
+    muted: 'bool',
+    volume: 'real',
+    playbackRate: 'real',
+    playbackState: 'enum', // MediaPlayer.StoppedState
```
akreuzkamp: Are you sure, that we can leave out the initialValue here?

src/modules/QtQuick/BorderImage.js
akreuzkamp: We should find a declarative solution for these as well, i.e. registering a type QuickBorder (or similar), which would be used here. But that can be done in another PR.

src/modules/QtQuick/Positioner.js
```diff
@@ -1,30 +1,28 @@
...
-            child.visibleChanged.connect(this, this.layoutChildren);
-    }
-}
+        for (const event of ['widthChanged', 'heightChanged', 'visibleChanged']) {
+          if (!child[event].isConnected(this, this.layoutChildren)) {
+            child[event].connect(this, this.layoutChildren);
+          }
+        }
+      }
+    });
```
akreuzkamp: What's the reason to put this change into this PR?

src/modules/QtQuick/RegExpValidator.js
```diff
@@ -3,11 +3,13 @@ registerQmlType({
...
+  properties: {
+    regExp: 'var'
+  }
+}, class {
+  constructor(meta) {
     callSuper(this, meta);
 
-    createProperty("var", this, "regExp");
-
     this.validate = (function(string) {
```
akreuzkamp: Isn't this function to be added as method as well?

src/modules/QtQuick/SystemPalette.js
```diff
@@ -18,17 +18,19 @@ registerQmlType({
...
+  constructor(meta) {
     callSuper(this, meta);
 
-    createProperty("enum", this, "colorGroup");
 
     var attrs    = [ 'alternateBase', 'base', 'button', 'buttonText', 'dark', 'highlight', 'highlightedText', 'light', 'mid', 'midlight', 'shadow', 'text', 'window', 'windowText' ];
     var platform = 'OSX';
 
     for (var i = 0 ; i < attrs.length ; ++i)
       createProperty("color", this, attrs[i], { readOnly: true });
```
akreuzkamp: We should add these explicitly, too.

src/modules/QtQuick/TextEdit.js
```diff
@@ -1,83 +1,68 @@
...
 
-    this.activeFocusOnPress = true;
     this.baseUrl = undefined;
-    this.canPaste = false;
-    this.canRedo = false;
-    this.canUndo = false;
-    this.color = 'white';
-    this.contentHeight = 0;
-    this.contentWidth = 0;
     this.cursorDelegate = undefined;
```
akreuzkamp: Why don't you morph these into initialValues?

src/modules/QtMultimedia/Video.js
ChALkeR: Yes, I made sure that `enum` defaults to 0.

src/modules/QtQuick/BorderImage.js
ChALkeR: Yes, there will be a solution for that, but I wanted to put that aside for this PR, because else it would become too huge =).

src/qtcore/qml.js
```diff
@@ -194,6 +194,10 @@ function callSuper(self, meta) {
     // TODO: not exported to the whole file scope yet
     Object.keys(info.enums).forEach(name => {
       self[name] = info.enums[name];
+
+      if (!global[name]) {
+        global[name] = self[name]; // HACK
+      }
```
akreuzkamp: Why?

src/modules/QtQuick/Positioner.js
ChALkeR: Hm, thanks. Looks like a mistake, will revert.

src/modules/QtQuick/RegExpValidator.js
ChALkeR: Yes, but that would require more code movement, and I wanted to do that in a follow-up PR.

src/modules/QtQuick/SystemPalette.js
ChALkeR: Ah, thanks. I missed that `attrs` list was hardcoded.
Will do.

akreuzkamp:
Looks mostly good to me! (see inline comments) :)

Though, I would prefer not to use class for now, to be honest (sorry for not having said that for #261 already). Instead I'd rather preserve the parameter `constructor` and add a parameter `methods`. This is mostly for consistency. Moreover it works with good old EcmaScript 5. But I'm ok with using class, too, if you decide to stick with it.

src/modules/QtQuick/Canvas.js
```diff
@@ -6,29 +6,22 @@ registerQmlType({
...
     baseClass: 'Item',
-    constructor: function(meta) {
+  properties: {
+    available: { type: 'bool', initialValue: true },
+    canvasSize: { type: 'var', initialValue: [0, 0] },
+    canvasWindow: { type: 'var', initialValue: [0, 0, 0, 0] },
+    context: { type: 'var', initialValue: {} },
+    contextType: { type: 'string', initialValue: 'contextType' },
+    renderStrategy: { type: 'enum', initialValue: 0 },
+    renderTarget: { type: 'enum', initialValue: 0 },
```
akreuzkamp: I think we can leave out these initialValues, right? :)

src/modules/QtQuick/TextEdit.js
ChALkeR: You mean the `undefined` variables? Will take a look at that, I don't remember atm why I omitted those.

src/modules/QtQuick/Canvas.js
ChALkeR: `renderStrategy` and `renderTarget`? Sure, will do.

akreuzkamp: yup, thanks :)

ChALkeR:
@akreuzkamp We don't need to bother with ES5 compatible syntax, as babel transplits things for us since cb24438. There are no signs of `class`es in the `lib` dir =).

Also, IMO we should prefer to use the features that are supported by spec instead of inveting our own syntax for that — that would make it _easier_ for someone to patch the sources, as there would be less unnecessary things to learn or care about. That's why I would prefer to not introduce something like a `methods` property without a need.

See also #263 (blocked).

Moreover, if there would have been a way to add typed properties directly to a class — I would probably have done that, instead. It might be possible that in future we would be able utilize something like the [ES Class Fields & Static Properties](https://github.com/jeffmo/es-class-fields-and-static-properties) and decorators for that.
It could look something like this:
```js
class {
  @qmlprop('int')
  myProp = 42;
}
```
Or not. Both those features are not finished yet.


We could still add support for `methods` property to make it easier to produce auto-generated code.

ChALkeR:
@akreuzkamp Btw, more stricter `class` syntax made me spot bugs like the one fixed in 472e11eb683f3eb801c886f0a9fc7593307b35e5.

That's doable without `class`, of course (those are actually just a syntax sugar), but would require a bit more code to insert our own safeguards.

akreuzkamp:
I'm all for #263. But I'm for a all-class-syntax or a all-initializer-object syntax, not some inconsistent syntax in between.

ChALkeR:
@akreuzkamp This is temporary until [ES Class Static Properties](https://github.com/jeffmo/es-class-fields-and-static-properties) get raised at least to stage-2, then we can go to «all-class» syntax, which will require only minor code changes.

Would that work?

Btw, I suppose we also should support «all-initializer-object» syntax, not for our own use, but purely for the auto-generated code (I suppose we should have a module compiler from QML at some point).

ChALkeR:
On a second thought, I would be fine with pulling #263 and reverting in the (relatively) unlikely case if it gets rejected or drastically changed.

@akreuzkamp, @Plaristote, what do you think?

src/modules/QtQuick/TextEdit.js
ChALkeR: Most of those were just wrong there. I just removed those, that would be better than it was before =).

It already behaves as `undefined` if `initialValue` is not specified, but those `undefined` are actually transformed into the type default values if they are specified for corresponding types. E.g. `selectedText` is a string, so it falls back to `''`, not `undefined`.

ChALkeR:
@akreuzkamp I resolved the issues you noticed, except for the mixed syntax for now.
Could you take a look again?

Mixed class/object syntax is temporary, I definitely want #263 merged at some point (btw, it has come coverage-related issues for now).

Anyway, even the mixed class/object syntax looks better to me, compared with what we have now (irregular prototype definitions, etc.). I will file PRs with more improvements (moving methods to the prototype, etc) after this gets landed :wink:.

stephenmdangelo:
It may be out of scope for this initial pass, but any thoughts on adding a "signals" argument as well?

ChALkeR:
@stephenmdangelo Yes, I am going to file a separate PR for that after this lands.

src/modules/QtQuick/Repeater.js
```diff
@@ -1,15 +1,23 @@
...
+  constructor(meta) {
     callSuper(this, meta);
     var self = this;
     var QMLListModel = getConstructor('QtQuick', '2.0', 'ListModel');
 
     this.parent = meta.parent; // TODO: some (all ?) of the components including Repeater needs to know own parent at creation time. Please consider this major change.
 
-    createProperty("Component", this, "delegate");
     this.container = function() { return this.parent; }
     this.$defaultProperty = "delegate";
```
akreuzkamp: You've overseen the default property here :)

akreuzkamp:
looks good to me now, let's merge it. :)
We can discuss the mixed class/object syntax later, if we need to :)

src/modules/QtQuick/Repeater.js
ChALkeR: Thanks, fixed!

PR #274, opened 2016-06-20
Revert passing in meta.context to createObject in construct

stephenmdangelo:
This was added in 2c7b240

By forcing the context for createObject in "construct" to that of the
calling context, it stops a new context Object from being created with
the Component's context as a prototype.  This will then cause changes
to the newly created Item's context to creep into the passed in
componentContext.

### DIFF
--- a/src/engine/qml.js
+++ b/src/engine/qml.js
@@ -247,6 +247,11 @@ function construct(meta) {
         // Currently we support only 1,2 and 4 and use order: 4,1,2
         // TODO: engine.qmldirs is global for all loaded components. That's not qml's original behaviour.
         var qdirInfo = engine.qmldirs[meta.object.$class]; // Are we have info on that component in some imported qmldir files?
+
+        /* This will also be set in applyProperties, but needs to be set here
+         * for Qt.createComponent to have the correct context. */
+        _executionContext = meta.context;
+
         if (qdirInfo) {
             // We have that component in some qmldir, load it from qmldir's url
             component = Qt.createComponent( "@" + qdirInfo.url);
--- a/src/engine/qml.js
+++ b/src/engine/qml.js
@@ -255,7 +255,7 @@ function construct(meta) {
             component = Qt.createComponent(meta.object.$class + ".qml");
 
         if (component) {
-            var item = component.createObject(meta.parent, {}, meta.context);
+            var item = component.createObject(meta.parent);
 
             if (typeof item.dom != 'undefined')
                 item.dom.className += " " + meta.object.$class + (meta.object.id ? " " + meta.object.id : "");
--- a/src/modules/QtQml/Component.js
+++ b/src/modules/QtQml/Component.js
@@ -18,18 +18,16 @@ QMLComponent.getAttachedObject = function() { // static
     return this.$Component;
 }
 
-QMLComponent.prototype.createObject = function(parent, properties, componentContext) {
+QMLComponent.prototype.createObject = function(parent, properties) {
     var oldState = engine.operationState;
     engine.operationState = QMLOperationState.Init;
     // change base path to current component base path
     var bp = engine.$basePath; engine.$basePath = this.$basePath ? this.$basePath : engine.$basePath;
 
-    if (!componentContext) componentContext = this.$context;
-
     var item = construct({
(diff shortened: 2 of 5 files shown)


stephenmdangelo:
Now, I don't know if @pavelvasev had a good reason for wanting to use the `meta.context` in `construct` instead of the Component's context as the base context.  Maybe the actual solution is to use the passed in `componentContext` in `Object.create` if it is passed in, instead of the Component's `$context`?

akreuzkamp:
+1 from me. 2c7b240 infact replaces standard behavior by non-standard behavior. Inline components are supposed to have the context of wherever they're defined and *not* where they're instanciated. The  same applies for components that are created using Qt.createComponent. The context is determined by the call to Qt.createComponent and not by the call to Component.createObject.
We should, though, wait for @pavelvasev to respond.

ChALkeR:
@akreuzkamp Actually, that change was reverted by @pavelvasev on his branch in cfb28a087285c68e5b76147fa1fe9440015bfc63. It also re-introduces passing `meta.context` to `Qt.createComponent`, which this PR doesn't do (it's useless now).

stephenmdangelo:
I have found a test-case that fails with this revert.  I've pushed it to the repo. I came across it while updating my use-case of qmlweb to use the latest master code.  I think there is some recent commit that now requires this change to work properly.  I will try and investigate to find what that is.

stephenmdangelo:
Yes, it's what I suspected re: `meta.context` vs `_executionContext`.  I've pushed a pretty hacky fix... any other ideas are welcome!

stephenmdangelo:
So, I'm thinking the best approach may be to create an internal `createComponent` method (maybe on QMLEngine) that takes a filename and a context.  This can be called from `Qt.createComponent` with `_executionContext` passed in, and from `construct` with `meta.context` passed in.

It can also be called from anywhere we currently call `new QMLComponent` (e.g. `loadQMLTree`, `createQmlObject`, `Loader.sourceChanged`, `Loader.sourceComponentChanged`), and will take care of setting the Component's `$basePath`, `$file` and `$imports` and calling `engine.loadImports` (this code is all duplicated in each of the aforementioned uses of `new QMLComponent`, except for `Loader`, which doesn't set `$basePath` or call `engine.loadImports`, breaking relative file paths).

I'll push some commits to this branch when I have them working, but I'm still open to other ideas.

ChALkeR:
So, what's going on here?

@akreuzkamp, does this look good to you in the current state?
If yes, let's just merge it — I don't think we should wait for @pavelvasev response here, given that that change was reverted in his branch =).

stephenmdangelo:
@ChALkeR I've rebased on master to solve the conflict.

> So, what's going on here?

The extra argument to `Qt.createComponent` was required because `_executionContext` is not always the context you want to pass on to the created `Component`.

I've solved the previously-failing test-case I added in this branch by just setting `_executionContext` around the call to `Qt.createComponent`.  Though I feel that my suggestion in https://github.com/qmlweb/qmlweb/pull/274#issuecomment-227746754 would be a cleaner solution.  I just wanted to get the go-ahead before diving into that level of refactoring.

src/engine/qml.js
```diff
@@ -234,15 +234,21 @@ function construct(meta) {
...
+        _executionContext = meta.context;
+
         if (qdirInfo) {
             // We have that component in some qmldir, load it from qmldir's url
             component = Qt.createComponent( "@" + qdirInfo.url);
         }
         else
             component = Qt.createComponent(meta.object.$class + ".qml");
 
+        _executionContext = oldExecutionContext;
```
akreuzkamp: I don't think that resetting the context here is required. Have you seen any errors, when omitting this? Instead we should just move setting `_executionContext` up, from `applyProperties` to here.

stephenmdangelo: @akreuzkamp, yes that should work fine. Though I think we need to keep line 365, as `applyProperties` is called manually by the QMLListElement constructor.

akreuzkamp:
lgtm :)

ChALkeR:
Merged in d7466f2ca92d17c55155e7438cf5b15c2c768a4b...f485359abc5dd8941def3b721be0846333606381, thanks!

PR #275, opened 2016-06-22
Component: set Item.parent in construct

stephenmdangelo:
This mirrors QML behaviour, which doesn't require the user to manually
set the item's parent after calling 'createObject'.

Currently, anywhere 'createObject' is called internally in qmlweb, the
item's 'parent' is manually set afterward, or the newly created items are
part of the default 'data' property, and 'QMLItem.dataChanged' does the
setting of the 'parent' property.

### DIFF
--- a/src/engine/qml.js
+++ b/src/engine/qml.js
@@ -262,6 +262,12 @@ function construct(meta) {
     // Apply properties (Bindings won't get evaluated, yet)
     applyProperties(meta.object, item, item, item.$context);
 
+    // Not checking if item inherits from Item because non-Item items may have
+    // a "parent" property (e.g. Timer).
+    if (meta.parent && item.hasOwnProperty("parent")) {
+      item.parent = meta.parent
+    }
+
     return item;
 }
 
--- a/src/modules/QtQuick/Item.js
+++ b/src/modules/QtQuick/Item.js
@@ -41,7 +41,10 @@ function QMLItem(meta) {
     this.dataChanged.connect(this, function(newData) {
         for (var i in newData) {
             var child = newData[i];
-            if (child.hasOwnProperty("parent")) // Seems to be an Item. TODO: Use real inheritance and ask using instanceof.
+
+            // Not checking if item inherits from Item because non-Item items
+            // may have a "parent" property (e.g. Timer).
+            if (child.hasOwnProperty("parent"))
                 child.parent = this; // This will also add it to children.
             else
                 this.resources.push(child);
--- a/tests/QMLEngine/basic.js
+++ b/tests/QMLEngine/basic.js
@@ -29,4 +29,11 @@ describe('QMLEngine.basic', function() {
     expect(qml.inner3).toBe(qml.current + 'foo/foo/lol/');
     expect(qml.full).toBe('http://example.com/bar');
   });
+
+  it('createObject', function() {
+    var qml = load('CreateObject', this.div);
+    expect(qml.children.length).toBe(1);
+    expect(qml.children[0].q).toBe(22);
+    expect(this.div.innerText).toBe("variable from context = 42");
+  });
 });
--- a/tests/QMLEngine/qml/BasicCreateObject.qml
+++ b/tests/QMLEngine/qml/BasicCreateObject.qml
@@ -1,0 +1,11 @@
+import QtQuick 2.2
+
+Item {
+  id: item
+  property string contextVar: "42";
+
+  Component.onCompleted: {
+    var c = Qt.createComponent("BasicCreateObjectSomeComponent.qml")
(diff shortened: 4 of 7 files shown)


src/modules/QtQml/Component.js
```diff
@@ -38,6 +38,10 @@ QMLComponent.prototype.createObject = function(parent, properties, componentCont
     engine.$basePath = bp;
 
     engine.operationState = oldState;
+
+    if (item.hasOwnProperty("parent") && parent) // Seems to be an Item. TODO: Use real inheritance and ask using instanceof.
```
stephenmdangelo: Note, the "TODO" comment here was copied from `QMLItem.dataChanged`. I'm not sure if there's now a better way to do this that I should use.

akreuzkamp: `item instanceof getConstructor('QtQuick.Item')` (where getConstructor needs to be replaced by whatever function you can get the constructor function of a class nowadays) should work by now. Would be cool, if you could verify that and use instanceof if it works. :)

akreuzkamp:
+1 :)

stephenmdangelo:
@akreuzkamp Yep, that worked, thanks.

If this is approved of, I can make a second commit that removes various places that item.parent is now set unnecessarily.

stephenmdangelo:
Bah, this messes up the order of children, so things in a Grid don't position properly.  Looking into it now...

stephenmdangelo:
I'm now setting the parent in `construct`, to ensure that items created with a constructor and those created with `createObject` still get their `parent` property set sequentially.  The first change I did caused children that were created with `createObject` to end up before children created with a constructor, as the latter weren't appended to `children` until `dataChanged`.

akreuzkamp:
nah, you should use `item.$properties.parent.set(parent, QMLProperty.ReasonInit);`. That should do the trick as well.

stephenmdangelo:
@akreuzkamp Sorry, are you suggesting putting `item.$properties.parent.set(parent, QMLProperty.ReasonInit);` into `Qt.createObject`?  This still results in the same problem, as the `parent` gets set for children created with `Qt.createObject` before it gets set for items created directly by `construct`.

stephenmdangelo:
I've rebased on master and kept it in `construct` for now.  I've also removed the check for inheritance from `Item`, as `Timer` also has a `parent` property (see https://github.com/qmlweb/qmlweb/pull/287).

src/engine/qml.js
```diff
@@ -262,6 +262,12 @@ function construct(meta) {
     // Apply properties (Bindings won't get evaluated, yet)
     applyProperties(meta.object, item, item, item.$context);
 
+    // Not checking if item inherits from Item because non-Item items may have
+    // a "parent" property (e.g. Timer).
```
akreuzkamp: Timer's parent property is not supposed to be set here. (See #287) ;)

src/modules/QtQuick/Item.js
```diff
@@ -41,7 +41,10 @@ function QMLItem(meta) {
     this.dataChanged.connect(this, function(newData) {
         for (var i in newData) {
             var child = newData[i];
-            if (child.hasOwnProperty("parent")) // Seems to be an Item. TODO: Use real inheritance and ask using instanceof.
+
+            // Not checking if item inherits from Item because non-Item items
+            // may have a "parent" property (e.g. Timer).
```
akreuzkamp: Same here. Timer's parent property is supposed to point to $parent, which already is set in QObject's constructor ;)

akreuzkamp:
I've looked at it in more detail now and in fact is not as easy as I thought. I made a review request with the approach, I would take to fix this issue. It's not perfect as well, but I'd prefer that: #296.

Btw.: Thank you for your contributions and sorry for being so nitpicky, all the time... I do appreciate your commits very much! :)

@ChALkeR Can you please have a look at this and #296 and choose your favorite? :)

stephenmdangelo:
@akreuzkamp no worries. No need to apologize for nitpicking, we all do it and we all just want the code to work! I don't take it personally. Also, I should have read my notifications NOT in reverse and would have seen all of that Timer parent stuff :)

ChALkeR:
@akreuzkamp 
> @ChALkeR Can you please have a look at this and #296 and choose your favorite? :)

Honestly, I have no idea :stuck_out_tongue_closed_eyes:. I can try to grok this a bit later, but I suppose it would be better if you sorted this out without me :smile:.

stephenmdangelo:
> I can try to grok this a bit later, but I suppose it would be better if you sorted this out without me :smile:.

Let's just go with #296.

stephenmdangelo:
Closing in favour of #296.

PR #276, opened 2016-06-22
Signal: Log out slot source on exception

stephenmdangelo:
This can help pin-point where the error is coming from


stephenmdangelo:
This is similar to the previous change I made for binding exceptions.

Plaristote:
LGTM, that should be helpful indeed.

Looks pretty weird when we see it in node's output, but considering the alternative is to not giving enough data to the user, I'd much rather have it that way than the previous one.

PR #279, opened 2016-07-03
gulpfile: separate coverage task from test task

ChALkeR:
This makes `npm test` much faster.

Coverage could be obtained by running `npm coverage`.


stephenmdangelo:
@ChALkeR Thank you! I can't +1 this one enough!

PR #285, opened 2016-07-03
Introduce QmlWeb variable, don't pollute global

ChALkeR:
Fixes: https://github.com/qmlweb/qmlweb/issues/51

We might want to add more methods/variables to the QmlWeb variable.
This does not deal with enums on a purpose, those should be covered by #271.

/cc @Plaristote @akreuzkamp @stephenmdangelo — PTAL.

### DIFF
--- a/src/qtcore/qmlstructure.js
+++ b/src/qtcore/qmlstructure.js
@@ -233,13 +233,13 @@ function convertToEngine(tree) {
 // Function to parse qml and output tree expected by engine
 function parseQML(src, file) {
     loadParser();
-    qmlweb_parse.nowParsingFile = file;
-    var parsetree = qmlweb_parse(src, qmlweb_parse.QmlDocument);
+    QmlWeb.parse.nowParsingFile = file;
+    var parsetree = QmlWeb.parse(src, QmlWeb.parse.QmlDocument);
     return convertToEngine(parsetree);
 }
 
 function loadParser() {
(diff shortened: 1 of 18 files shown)


src/qtcore/qmlstructure.js
```diff
@@ -255,6 +255,9 @@ function loadParser() {
           throw new Error('Could not load QmlWeb parser!');
       }
       (new Function(xhr.responseText))();
+      // TODO: remove in 0.2, qmlweb.parser should set to QmlWeb if QmlWeb is defined
+      QmlWeb.qmlweb_parse = QmlWeb.qmlweb_parse || qmlweb_parse;
+      QmlWeb.qmlweb_jsparse = QmlWeb.qmlweb_jsparse || qmlweb_jsparse;
```
akreuzkamp: What do you think about renaming that to `QmlWeb.parse` and `QmlWeb.parseJs`? :)

akreuzkamp:
lgtm :)

src/qtcore/qmlstructure.js
ChALkeR: Will do, thanks! =)

ChALkeR:
Merged.

Corresponding parser change: https://github.com/qmlweb/qmlweb-parser/commit/bb2aeaa45a8d69b95d3e4d85e3e1a7fd10d8381f

PR #287, opened 2016-07-04
Timer: Added parent property

stephenmdangelo:
C++ QML has the parent property explicitly defined for Timer components,
since Timer doesn't inherit from Item:

https://github.com/qt/qtdeclarative/blob/5.7.0/src/qml/types/qqmltimer_p.h#L72

### DIFF
--- a/src/modules/QtQml/Timer.js
+++ b/src/modules/QtQml/Timer.js
@@ -5,6 +5,7 @@ registerQmlType({
   baseClass: "QtObject",
   properties: {
     interval: { type: "int", initialValue: 1000 },
+    parent: { type: "QtObject", readOnly: true },
     repeat: "bool",
     running: "bool",
     triggeredOnStart: "bool"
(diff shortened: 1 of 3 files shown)


stephenmdangelo:
Probably want to wait for #271 to land for this one, but I used `properties` to make it easier to merge after that.

ChALkeR:
LGTM.

Just in case: note that we should't be copying or using Qt sources due to incompatible licenses.
This change is trivial, though, but could you at least remove the link from the commit? :-)

stephenmdangelo:
>  This change is trivial, though, but could you at least remove the link from the commit? :-)

Done

akreuzkamp:
lgtm :)

akreuzkamp:
I checked what Timer does, again and recognized that it differs from what this patch does. The confusing thing here is, that QML has to different concepts of parents:
* A logical parent, this is an (internal) property of QObject and in QmlWeb is called $parent. 
* A graphical parent, this is what Item has.

What timer does, is exposing the *logical* parent ($parent) to QML. Thus to imitate that behavior, we need a readonly property, whose getter returns $parent.

stephenmdangelo:
@akreuzkamp Should be good now.

stephenmdangelo:
Rebased. Any objections?

ChALkeR:
Nope. Merging =). Thanks!

PR #288, opened 2016-07-04
Export callSuper and createProperty via QmlWeb

stephenmdangelo:
This allows new custom QML Components to be defined in .js files outside
of lib/qt.js.


ChALkeR:
Could you move those to the bottom, like all the other exports?

I would prefer to get rid of `callSuper` somehow, but exporting it should work for now.

stephenmdangelo:
> Could you move those to the bottom, like all the other exports?

Done

> I would prefer to get rid of callSuper somehow, but exporting it should work for now.

Yeah, we had briefly discussed it here: https://github.com/qmlweb/qmlweb/issues/260#issuecomment-225452775

ChALkeR:
LGTM, merging.

PR #290, opened 2016-07-05
JSItemModel: initial cleanup

ChALkeR:
* Convert JSItemModel to class.
* Move JSItemModel documentation to `docs/` folder.

/cc @akreuzkamp @Plaristote 

This PR creates a `docs/` folder =).

### DIFF
--- a/docs/JSItemModel.md
+++ b/docs/JSItemModel.md
@@ -1,0 +1,31 @@
+# JSItemModel
+
+JSItemModel is a QAbstractItemModel like class that allows to create models
+from javascript, that can later be used from QML.
+
+## To implement a model
+
+* Create a `JSItemModel` object and add `rowCount()` and `data()` functions to
+  the object,
+* use `setRoleNames` to define `roleNames`,
+* emit `dataChanged`, `rowsInserted`, `rowsMoved`, `rowsRemoved` and
+  `modelReset` as your internal data changes.
+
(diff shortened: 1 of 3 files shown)


Plaristote:
lgtm

akreuzkamp:
Are you sure, we want to move the documentation to a separated file?
In the C++ world, it's considered a feature if the documentation is inside the code. You then use a tool like [doxygen](https://en.wikipedia.org/wiki/Doxygen), to create a documentation out of it.

ChALkeR:
@akreuzkamp That's how it's done usually.

In the C++ world, headers are separate from the implementation, and headers contain the documentation that also serves for autocompletion purposes.

We could use [JSDoc](http://usejsdoc.org/) format, for example, but I'm not sure if that would help anyone — I doubt that autocompletion would work for most setups, or that many users would be glad to check the comments in the code manually. Also, excessively detailed documentation in the `*.js` files would also make our _actual implementation_ files larger and harder to gasp for the developers, which is something I want to avoid.

Keeping the documentation accessible is important, and placing it in the `docs/` dir gives us that with no additional burden — you can open the `docs/` section for any branch or any release tag. See https://github.com/qmlweb/qmlweb/blob/master/docs/JSItemModel.md, for example.

I'm open for discussion, though, but I suppose that should be done in a separate issue =).

PR #293, opened 2016-07-05
QtQml/Qt: Added Qt.size method

stephenmdangelo:


ChALkeR:
Hm. I don't think we should add manual creation of those objects to the `Qt` impl. Perhaps `QSizeF`/`QRectF`/`QPointF` should be added as separate classes to the `src/qtbase` dir and only created in `Qt` as `return new QSizeF(width, height)`?

akreuzkamp:
+1 to @ChALkeR. I wanted to comment exactly the same thing :)

gulpfile.js
```diff
@@ -20,6 +20,7 @@ const qtcoreSources = [
   'src/engine/import.js',
   'src/engine/qml.js',
   'src/engine/*.js',
+  'src/qtbase/QObject.js',
```
stephenmdangelo: @ChALkeR @akreuzkamp I assumed this was a good idea to ensure that QObject.js is included before other files that may try to inherit from it (like QSizeF)? Just in case QObject turns into a class and has to be declared first.

src/qtbase/QSizeF.js
```diff
@@ -0,0 +1,7 @@
+class QSizeF extends QObject {
```
stephenmdangelo: This is my first time using "real" JavaScript class inheritance, so please let me know if there are any issues.

ChALkeR: No, should be good. Btw, that's not the «real» class inheritance, it's the same prototypes with a syntax sugar =).

src/qtbase/QSizeF.js
```diff
@@ -0,0 +1,7 @@
...
+  constructor( width, height ) {
+    super();
+    createProperty("real", this, "width", { initialValue : width });
```
ChALkeR: Single (`'`) quotes are slightly more common in the sources than double (`"`). I am not yet sure if we would want to unify them to a single format (i.e. single vs double), but in case if we do, I don't see a reason for adding the new code with the less popular one =).

Could you replace the quotes in the newly added js files to single (`'`), please?

src/qtbase/QSizeF.js
```diff
@@ -0,0 +1,7 @@
+class QSizeF extends QObject {
+  constructor( width, height ) {
```
ChALkeR: Lint: `constructor( width, height ) {`.

ChALkeR: Lint: `constructor(width, height) {`.

src/qtbase/QSizeF.js
```diff
@@ -0,0 +1,7 @@
...
+    super();
+    createProperty("real", this, "width", { initialValue : width });
+    createProperty("real", this, "height", { initialValue : height });
```
ChALkeR: Lint: `initialValue: height` (note no space after the `initialValue`).

gulpfile.js
ChALkeR: Yes, makes perfect sense.

ChALkeR:
LGTM with minor nits.

ChALkeR:
Lint tests are not yet enabled on the `src/` dir, but I expect at least the new files not to cause any new lint errors without a good reason =).

src/qtbase/QSizeF.js
stephenmdangelo: Well, I _did_ put "real" in quote :)

stephenmdangelo:
Should be good now, I've also now figured out how to run `eslint` :+1:

src/qtbase/QSizeF.js
ChALkeR: Ah, missed that :smile:

ChALkeR:
Merged in 957ba0d666125bc765e8a291d5c9c3e48e2fa664...b2e66fe990723b198066da25023864d39ade69bf.

PR #296, opened 2016-07-09
Component: set Item.parent in createObject - alternative

akreuzkamp:
This is an alternative approach to the issue, that #275 tries to fix.
This approach isn't very nice, either, but I consider it at least a bit better.

The issue I see with the other approach is that it sets `parent` (including all its consequences with inserting it into the dom, etc.) in `construct` just to override it later, in case it wasn't supposed to.

### DIFF
--- a/src/engine/QMLEngine.js
+++ b/src/engine/QMLEngine.js
@@ -176,7 +176,7 @@ class QMLEngine {
     component.$imports = tree.$imports; // for later use
     component.$file = file; // just for debugging
 
-    this.rootObject = component.createObject(parentComponent);
+    this.rootObject = component.$createObject(parentComponent);
     component.finalizeImports(this.rootContext());
     this.$initializePropertyBindings();
 
--- a/src/engine/qml.js
+++ b/src/engine/qml.js
@@ -252,7 +252,7 @@ function construct(meta) {
             component = Qt.createComponent(meta.object.$class + ".qml");
 
         if (component) {
-            var item = component.createObject(meta.parent);
+            var item = component.$createObject(meta.parent);
 
             if (typeof item.dom != 'undefined')
                 item.dom.className += " " + meta.object.$class + (meta.object.id ? " " + meta.object.id : "");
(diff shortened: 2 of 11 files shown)


ChALkeR:
@stephenmdangelo does this look good to you?

stephenmdangelo:
@ChALkeR yep. Probably want to squash the two commits since some of the initial commit was removed in the second one. I'm good with not getting credit for the tests ;)

src/modules/QtQuick/Item.js
```diff
@@ -41,7 +41,10 @@ function QMLItem(meta) {
     this.dataChanged.connect(this, function(newData) {
         for (var i in newData) {
             var child = newData[i];
-            if (child.hasOwnProperty("parent")) // Seems to be an Item. TODO: Use real inheritance and ask using instanceof.
+
```
stephenmdangelo: @ChALkeR this change is not necessary, as Timer has a different parent as mentioned in #287.

ChALkeR: But there seems to be no change here, except for the comments?

stephenmdangelo: Yeah, but any reference to setting the Timer parent I wrote in these comments is now misleading as the Timer parent doesn't have to be set here and is handled on its own.

ChALkeR: The change in `src/engine/qml.js` is still needed, though?

stephenmdangelo: Nope, it's removed in the next commit.

ChALkeR: Done, PTAL.

ChALkeR:
@stephenmdangelo @akreuzkamp Partially squashed, rebased, reordered, amended comments.

Could you take a look?

ChALkeR:
Updated once again, fixed issues after rebase.

stephenmdangelo:
Looks good to me.

ChALkeR:
LGTM, landing.

PR #299, opened 2016-07-10
Qt, qtbase: add more basic types constructors

ChALkeR:
~~Note: this adds just simple constructors for them, methods are not there yet.~~
~~This also doesn't property types for those yet.~~

/cc @akreuzkamp @stephenmdangelo


labsin:
They belong in the QtQuick namespace iso QtQml:
doc.qt.io/qt-5/qtquick-qmlmodule.html

Do they also work as basic types? (property point a: "1,2")

Also:
"Be aware, however, that a property change signal for a basic type is
emitted whenever *any* of its attributes have changed, as well as when the
property itself changes. Take the following code, for example:"

ChALkeR:
@labsin Thanks!

About the placement of those — those are not regular Qt types, but basic types. Sure, they are provided by the QtQml, but are not instanceable by their names from qml (e.g. QVector2D), and what QtQml exposes looks like wrappers around (or bindings to) standard e.g. QVector2D, which is a part of Qt Gui, not Qt QML.

This is why the actual impl is placed to a separate directory, and QtQml namespace should just expose those — currently it does that via `Qt.vector2d` etc. (`Qt` is also a part of QtQml).

> Do they also work as basic types? (property point a: "1,2")

Not yet, this just adds basic constructors for now and means to access them through the `Qt` variable from QtQml. Actually making them usable as a first-class qml basic types would require more work, and should be done separately. I altered the commit message to reflect that.

I will open an issue for that as soon as this lands.

> Be aware, however, that a property change signal for a basic type is
emitted whenever *any* of its attributes have changed, as well as when the
property itself changes

I suppose that requires special support from the QMLProperty side. That's not supported atm, I will open an issue for that as soon as this lands.

labsin:
Op 10 jul. 2016 11:23 schreef "Сковорода Никита Андреевич" <
notifications@github.com>:
>
> @labsin Thanks!
>
> About the placement of those — those are not regular Qt types, but basic
types. Sure, they are provided by the QtQml, but are not instanceable by
their names from qml (e.g. QVector2D), and what QtQml exposes looks like
wrappers around (or bindings to) standard e.g. QVector2D, which is a part
of Qt Gui, not Qt QML.
>
> This is why the actual impl is placed to a separate directory, and QtQml
namespace should just expose those — currently it does that via Qt.vector2d
etc. (Qt is also a part of QtQml).

The docs say the basic type wrapper are part of QtQuick but the Qt global
object is part of QtQml. And the Qt object has methods that return the
basic types. It all seems a bit weird. They should indeed be included in
the base but we must make sure the aliased imports work.

>>
>> Do they also work as basic types? (property point a: "1,2")
>
> Not yet, this just adds basic constructors for now and means to access
them through the Qt variable from QtQml. Actually making them usable as a
first-class qml basic types would require more work, and should be done
separately. I altered the commit message to reflect that.
>
> I will open an issue for that as soon as this lands.
>>
>> Be aware, however, that a property change signal for a basic type is
>> emitted whenever any of its attributes have changed, as well as when the
>> property itself changes
>
> I suppose that requires special support from the QMLProperty side. That's
not supported atm, I will open an issue for that as soon as this lands.

src/qtbase/QObject.js
```diff
@@ -40,3 +40,7 @@ function QObject(parent) {
     // http://doc.qt.io/qt-5/qtqml-javascript-dynamicobjectcreation.html
     this.destroy = this.$delete;
 }
+
+QObject.prototype.toString = function() {
+    return `${this.constructor.name}()`;
```
stephenmdangelo: Maybe add a `QObject.$toString(s)`, where "s" is put in between the parentheses? Save all the calls to "replace"?

ChALkeR: Makes sense, thanks! Will do.

ChALkeR: Done! PTAL

ChALkeR:
@labsin In order for `property point a: "1,2"` and so on to work, #307 should be landed first — that includes fixes for classes vs plain function wrappers for property creation. After that one lands, I will either update this or file a new pull request to support defining properties of the types listed here and setting them using strings (where appropriate).

ChALkeR:
> The docs say the basic type wrapper are part of QtQuick

Yes, that is correct. More explicitly, i.e. `QMLColor` should be a wrapper around `QColor` that stores the `QColor` instance inside it, but does not inherit it. Ditto for all the rects/sizes/etc.

But do we really want to overcomplicate things? The only thing that could be needed for is removing (hiding) some methods/properties from e.g. `QMLColor` which were present in `QColor`.  I.e. `QMLColor`  has only these properties: `toString`, `constructor`, `toLocaleString`, `valueOf`, `hasOwnProperty`, `isPrototypeOf`, `propertyIsEnumerable`, `__defineGetter__`, `__defineSetter__`, `__proto__`, while `QColor` should perhaps have more.

ChALkeR:
This should mostly work now.

Property types for those should be supported, and `onPropertyChanged` should be fired with the basic type is modified by changing its child properties.

ChALkeR:
> "Be aware, however, that a property change signal for a basic type is
emitted whenever any of its attributes have changed, as well as when the
property itself changes. Take the following code, for example:"

This is now implemented.

PR #300, opened 2016-07-10
Signal: convert to class and cleanup

ChALkeR:
Docs are currently mostly missing, but I want to merge this as-is, because I don't want to write documentation for those methods atm :smile:.

/cc @akreuzkamp


ChALkeR:
@akreuzkamp Going to merge this today if no objections.

akreuzkamp:
lgtm :)

PR #301, opened 2016-07-10
Automatic signal handlers

ChALkeR:
/cc @akreuzkamp @Plaristote, could you take a look?

~~This adds some new methods called `on{Something}` to several types, the idea here is to catch and bind those automatically to corresponding signals in `callSuper` later.~~

~~**Update:** second commit here uses those methods to automatically connect signal handlers.~~

~~Is `on{Something}` fine? Would this work without conficts when a user has their `on{Something}` defined? Should I rename those to `$on{Something}`? I have not checked =).~~

This uses the already added `on{Something}` methods to automatically connect signal handlers.

### DIFF
--- a/src/engine/modules.js
+++ b/src/engine/modules.js
@@ -174,8 +174,11 @@ function inherit(constructor, baseClass) {
 
 function callSuper(self, meta) {
   const info = meta.super.$qmlTypeInfo || {};
+  meta.superDepth = meta.superDepth || 0;
+  meta.superDepth++;
   meta.super = meta.super.prototype.constructor;
   meta.super.call(self, meta);
+  meta.superDepth--;
 
   if (info.enums) {
     // TODO: not exported to the whole file scope yet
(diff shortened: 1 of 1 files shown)


akreuzkamp:
I guess it would cause no conflicts. Anyway I would not risk to break things, if we change that later. So we should definitely go for $onXChanged.
Anyway, I'm not really convinced about this. We need to supersede our current approach to react to property changes. But I think we need to get away from connecting to the public xChanged-signal completely and move towards setters. I'm not sure, how the final solution would look like, though.

ChALkeR:
> I guess it would cause no conflicts. Anyway I would not risk to break things, if we change that later. So we should definitely go for $onXChanged.

Will do, thanks.

> We need to supersede our current approach to react to property changes. But I think we need to get away from connecting to the public xChanged-signal completely and move towards setters. I'm not sure, how the final solution would look like, though.

Perhaps, but that should be done on by-file basis. The main idea of this change here is to keep the code cleaner and more maintainable without changing the actual behavior.

I believe this would in fact simplify future rewrite to whatever approach =).

Btw, this still requires some improvements, perhaps I should cache the matched methods in the type descriptor, and perhaps I should add a field that would allow the compiler to pass an already populated list of matched connections to avoid matching those in the runtime.

akreuzkamp:
ok. Then I'd still rather call these changed handlers explicitly from `QMLProperty.set` than doing this connect magic.
I'd keep those signal handlers, which aren't propertyChanged handlers as they are for now (is is more than `Component$onCompleted` anyway?).

ChALkeR:
> Then I'd still rather call these changed handlers explicitly from QMLProperty.set than doing this connect magic.

You mean writing setters for each property? The thing is — it's not easy to call all those — they might be deep, like `foo$bar$onSomethingChanged` — with only `bar` being the immediate parent of the `something` property here. Is your concern here that the signals are slower than calling those directly or that the handlers should be called before saving the actual value for some cases?

> (is is more than Component$onCompleted anyway?).

Yes, even this PR has `Keys$onPressed`.

akreuzkamp:
Concerning Keys$onPressed: We should keep this one as it is for now. Connecting to it from Item is wrong anyway (breaks modularity) and I'm just preparing a merge of my keyhandling code from the realInheritance branch.

My concern is, that the code obscures how the connection is done and doing it `callSuper` is as well hidden as it can get...

ChALkeR:
@akreuzkamp I will probably split this into a few independant pull requests, so we could discuss things separately =).

ChALkeR:
I'm not sure now if this is the way forward, but I still keep the milestone as this is something I want to be resolved rather sooner than later.

ChALkeR:
Yeah, on a second thought, this doesn't look like a good idea and doesn't seem to map nicely to JS.
Closing.

PR #302, opened 2016-07-11
engine/qml: Fix typo when creating declarative signals

stephenmdangelo:
This was causing parameters in signals to not be accessible in slots.


ChALkeR:
Argh! LGTM, thanks!

PR #304, opened 2016-07-14
Fix updateH/Vgeometry recursion from implicitWidth/Height changes

stephenmdangelo:
The first commit is not required for this, and I don't have an explicit test case for it, but I think it just makes sense based on the code paths... @akreuzkamp Thoughts?

The second commit is the real fix for the test case that was modified.


akreuzkamp:
regarding the first commit: mhh.. We only need that connection if we are using x and y, that's why it is done there, right now. Could it be that changing this, would cause a lot of unneeded geometry updates?

regarding the second commit: lgtm :)

stephenmdangelo:
> regarding the first commit: mhh.. We only need that connection if we are using x and y, that's why it is done there, right now. Could it be that changing this, would cause a lot of unneeded geometry updates?

@akreuzkamp You are correct. I was just seeing that it was using the property in all of the other code paths, so assumed that it may need the connection, but those other code paths are just updating the "x" value as a result of anchoring, and no change in the parent's "left" value will cause the relative "x" of a child to change.

I've removed it.

stephenmdangelo:
Any objections to this one?

ChALkeR:
Am I correct that the only commit left here was reviewed by @akreuzkamp?

stephenmdangelo:
That's right. I removed the unneeded commit.

ChALkeR:
Merged, thanks!

Issue #283, opened 2016-07-03
Wrong text wrapping with anchors.horizontalCenter alignment


ChALkeR:
```qml
import QtQuick 2.0

Rectangle {
  implicitWidth: 300
  implicitHeight: 100
  color: 'lightgray'

  Text {
    text: 'Hello world!'
    y: 10
    anchors.horizontalCenter: parent.horizontalCenter
    font.pointSize: 24
    font.bold: true
  }
}
```

The text is wrapped to two lines for some reason, while with `x: 0` instead of `anchors.horizontalCenter` it's displayed as a single line.

ChALkeR:
Fixed by https://github.com/qmlweb/qmlweb/pull/305.

PR #305, opened 2016-07-15
QtQuick/Text: Move span to document body before calculating implicit size

stephenmdangelo:
This is required as, once the size is set once, it will not be properly
re-calculated when any other property connected to "updateImplicit" is
changed, as the offsetWidth/offsetHeight will always return the current
size.

### DIFF
--- a/src/modules/QtQuick/Text.js
+++ b/src/modules/QtQuick/Text.js
@@ -116,6 +116,7 @@ registerQmlType({
     this.text = "";
 
     this.textChanged.connect(this, updateImplicit);
+    this.wrapModeChanged.connect(this, updateImplicit);
     this.font.boldChanged.connect(this, updateImplicit);
     this.font.weightChanged.connect(this, updateImplicit);
     this.font.pixelSizeChanged.connect(this, updateImplicit);
@@ -128,10 +129,22 @@ registerQmlType({
 
     function updateImplicit() {
         if (typeof this.text == undefined || this.text === "" || !this.dom) {
-             this.implicitHeigh = this.implicitWidth = 0;
+             this.implicitHeight = this.implicitWidth = 0;
         } else {
-            this.implicitHeight = fc.offsetHeight;
-            this.implicitWidth = fc.offsetWidth;
+            /* Need to move the child out of it's parent so that it can
+             * properly recalculate it's "natural" offsetWidth/offsetHeight
+             * */
+            if (this.$isUsingImplicitWidth) {
+                document.body.appendChild(fc);
+            }
+            var fc_offsetHeight = fc.offsetHeight;
+            var fc_offsetWidth = fc.offsetWidth;
+            if (this.$isUsingImplicitWidth) {
+                this.dom.appendChild(fc);
+            }
+
+            this.implicitHeight = fc_offsetHeight;
+            this.implicitWidth = fc_offsetWidth;
         }
     }
   }
--- a/tests/QtQuick/Text.js
+++ b/tests/QtQuick/Text.js
@@ -1,0 +1,10 @@
+describe("QtQuick.Text", function() {
+  setupDivElement();
+
+  var load = prefixedQmlLoader("QtQuick/qml/Text");
+  it("implicit size", function() {
+    var qml = load("ImplicitSize", this.div);
+    expect(qml.text_item.width).toBeGreaterThan(0);
+  });
+});
+
--- a/tests/QtQuick/qml/TextImplicitSize.qml
+++ b/tests/QtQuick/qml/TextImplicitSize.qml
@@ -1,0 +1,11 @@
+import QtQuick 2.0
+
+Item {
+  property alias text_item: _text_item
+  Rectangle {
+    Text {
+      id: _text_item
+      text: "_ _ _"
+    }
+  }
+}


stephenmdangelo:
I wasn't able to get a Render test to work properly. I'm assuming it's the differing font rendering in Qt vs the test browser, but the test case here catches the previous error.

Moving the span DOM element feels a little gross, so I'm open to any other solution to this problem.

ChALkeR:
Does this fix #283?

ChALkeR:
As for the implementation: I am not sure if moving the current element is a good idea or not. Perhaps we should implement QML [TextMetrics](http://doc.qt.io/qt-5/qml-qtquick-textmetrics.html) and use that instead.

`TextMetrics` could create a single (shared) hidden element and change its content and style to recalculate preferred dimensions for each given text.

Plaristote:
Quick note : going over that, it seems that lines 112/113 don't have any purpose anymore in Text.js.

TextMetrics also requires some work with the ellipsis styling, but that doesn't seem to be necessary right now. Should we start by implementing the features that `Text` already has in a new TextMetrics object ?
Maybe that could go in this PR, and then we'll finish the implementation of TextMetrics in due time.

stephenmdangelo:
@ChALkeR 
> Does this fix #283?

Yes, it does.

> TextMetrics could create a single (shared) hidden element and change its content and style to recalculate preferred dimensions for each given text.

I did some quick profiling in Firefox:

```
        function updateImplicit() {
            if (_typeof(this.text) == undefined || this.text === "" || !this.dom) {
                this.implicitHeigh = this.implicitWidth = 0;
            } else {
                /* Need to move the child out of it's parent so that it can
                 * properly recalculate it's "natural" offsetWidth/offsetHeight
                 * */

              console.time("move span")
              for(var i=0; i<1000; ++i) {
                document.body.appendChild(fc);
                var fc_offsetHeight = fc.offsetHeight;
                var fc_offsetWidth = fc.offsetWidth;
                this.dom.appendChild(fc);
              }
              console.timeEnd("move span")

              console.time("transfer properties cssText")
              for(var i=0; i<1000; ++i) {
                QmlWeb.$textMetricsItem.style.cssText = fc.style.cssText;
                QmlWeb.$textMetricsItem.innerHTML = fc.innerHTML;
                var fc_offsetHeight = QmlWeb.$textMetricsItem.offsetHeight;
                var fc_offsetWidth = QmlWeb.$textMetricsItem.offsetWidth;
                QmlWeb.$textMetricsItem.style.display = 'none';
                QmlWeb.$textMetricsItem.innerHTML = "";
              }
              console.timeEnd("transfer properties cssText")

              console.time("transfer properties")
              for(var i=0; i<1000; ++i) {
                QmlWeb.$textMetricsItem.style.fontStyle = fc.style.fontStyle;
                QmlWeb.$textMetricsItem.style.fontVariant = fc.style.fontVariant;
                QmlWeb.$textMetricsItem.style.fontWeight = fc.style.fontWeight;
                QmlWeb.$textMetricsItem.style.fontSize = fc.style.fontSize;
                QmlWeb.$textMetricsItem.style.fontFamily = fc.style.fontFamily;
                QmlWeb.$textMetricsItem.style.lineHeight = fc.style.lineHeight;
                QmlWeb.$textMetricsItem.innerHTML = fc.innerHTML;
                QmlWeb.$textMetricsItem.style.display = null;
                var fc_offsetHeight = QmlWeb.$textMetricsItem.offsetHeight;
                var fc_offsetWidth = QmlWeb.$textMetricsItem.offsetWidth;
                QmlWeb.$textMetricsItem.style.display = 'none';
                QmlWeb.$textMetricsItem.innerHTML = "";
              }
              console.timeEnd("transfer properties")

              console.time("canvas")
              for(var i=0; i<1000; ++i) {
                var fc_offsetHeight = this.font.pixelSize;
                var the_font = fc.style.fontStyle + " " + fc.style.fontVariant + " " + fc.style.fontWeight + " " + fc.style.fontSize + /*"/" + fc.style.lineHeight +*/ " " + fc.style.fontFamily;
                QmlWeb.$textMetricsContext.font = the_font;
                var t = QmlWeb.$textMetricsContext.measureText(this.text)
                var fc_offsetWidth = t.width;
              }
              console.timeEnd("canvas")

                console.log("setting implicitWidth", fc_offsetWidth, fc.offsetWidth, QmlWeb.$textMetricsContext.font)

                this.implicitHeight = fc_offsetHeight;
                this.implicitWidth = fc_offsetWidth;
            }
        }
```

Output:

```
move span: timer started
move span: 109.26ms
transfer properties cssText: timer started
transfer properties cssText: 169.28ms
transfer properties: timer started
transfer properties: 224.48ms
canvas: timer started
canvas: 48.4ms
```

The issue with the `canvas` method is that it doesn't calculate the `implicitHeight` properly.  A method I found to find the height involves rendering to an image and scanning the pixels of that image.  I'm going to assume that is slower: http://pomax.github.io/fontmetrics.js/fontmetrics.js.

Also, my current method does not properly calculate JUST the `implicitHeight` in the case that the width is explicitly set and wrapping is turned on.

ChALkeR:
@stephenmdangelo I have also considered the `canvas`-based impl, but that wouldn't work for anything that wraps lines or has some complex markup. Note that `Text` could have rich formatting.

Let's put `canvas` aside for now =). Perhaps it could be used for optimization of corner cases (when width is unbounded, rich text is disabled, etc), but not for the generic case.

About other methods — do they (in their current impl) support limiting the `width`, for example?

It looks like re-attaching the span is faster. That would be fine by me, as long as it doesn't cause any flickering (which I believe it doesn't). :smile:

stephenmdangelo:
I'll take a look at handling explicit `width` setting.

Just to round out the tests, I tested with the `$textMetricsItem` being detached from the DOM when making changes, and that was better, but still worse than moving:

```
move span: timer started
move span: 110.26ms
transfer properties cssText: timer started
transfer properties cssText: 144.75ms
transfer properties: timer started
transfer properties: 161.02ms
```

stephenmdangelo:
> About other methods — do they (in their current impl) support limiting the width, for example?

Done. Now this example looks a whole lot like `qmlscene`:

```
import QtQuick 2.0

Column {
  spacing: 2
  Rectangle {
    implicitWidth: 300
    implicitHeight: 100
    color: 'lightgray'

    Rectangle {
      color: "red"
      anchors.fill: text_item1
    }

    Text {
      id: text_item1
      text: 'Hello world!'
      width: 10
      anchors.centerIn: parent
      font.pointSize: 24
      font.bold: true
      wrapMode: Text.WordWrap
      Component.onCompleted: {
        wrapMode = Text.NoWrap
      }
    }
  }
  Rectangle {
    implicitWidth: 300
    implicitHeight: 100
    color: 'lightgray'

    Rectangle {
      color: "red"
      anchors.fill: text_item2
    }

    Text {
      id: text_item2
      text: 'Hello world!'
      width: 10
      anchors.centerIn: parent
      font.pointSize: 24
      font.bold: true
      wrapMode: Text.WordWrap
      Component.onCompleted: {
        wrapMode = Text.NoWrap
        wrapMode = Text.WordWrap
      }
    }
  }
  Rectangle {
    implicitWidth: 300
    implicitHeight: 100
    color: 'lightgray'

    Rectangle {
      color: "red"
      anchors.fill: text_item3
    }

    Text {
      id: text_item3
      text: 'Hello world!'
      width: 10
      height: 10
      anchors.centerIn: parent
      font.pointSize: 24
      font.bold: true
      wrapMode: Text.WordWrap
      Component.onCompleted: {
        wrapMode = Text.NoWrap
        wrapMode = Text.WordWrap
      }
    }
  }
  Rectangle {
    implicitWidth: 300
    implicitHeight: 100
    color: 'lightgray'

    Rectangle {
      color: "red"
      anchors.fill: text_item4
    }

    Text {
      id: text_item4
      text: 'Hello world!'
      width: 10
      height: 10
      anchors.centerIn: parent
      font.pointSize: 24
      font.bold: true
      wrapMode: Text.WordWrap
      Component.onCompleted: {
        wrapMode = Text.NoWrap
      }
    }
  }
  Rectangle {
    implicitWidth: 300
    implicitHeight: 100
    color: 'lightgray'

    Rectangle {
      color: "red"
      anchors.fill: text_item5
    }

    Text {
      id: text_item5
      text: 'Hello world!'
      anchors.centerIn: parent
      font.pointSize: 24
      font.bold: true
      wrapMode: Text.WordWrap
      Component.onCompleted: {
        wrapMode = Text.NoWrap
      }
    }
  }
}
```

This fixes cases 1 and 5. Case 2 I had broken with my initial fix.

> It looks like re-attaching the span is faster. That would be fine by me, as long as it doesn't cause any flickering (which I believe it doesn't). :smile: 

I've not noticed any.

Another issue I've noticed is that we aren't always properly setting initial CSS styles (e.g. we should be calling the "changed" handler for most things that apply CSS styles during initialization, as the CSS default often does not match the QML default).  I had made a fix for Scale (a6628d68be53161a0fa3a6e9d493e3fd97267238), and have noticed that many things in `Font.js` have the same issue (that's why I have to explicitly set `Text.NoWrap` in my above example).

I can go through and manually add calls in `Font.js` to all of the `changed` handlers in the constructor, but I feel like that's a thing that could be handled with #301.

@Plaristote, in regards to your comment about lines 112/113 (e.g. setting of the initial font family and point size), those don't actually seem to get propagated to the element's style anyway, since they haven't "changed" from the defaults in `Font.js`.

ChALkeR:
LGTM, landing.

PR #306, opened 2016-07-16
modules: cleanup some qml types, merge geometry into Item

ChALkeR:
I tried to keep actual code changes to the minimum, this is mostly code movement.

Not all types completed, this is about ½ of them =).

@akreuzkamp, @Plaristote PTAL.


ChALkeR:
Note to self: should not forget about applying #304 if that lands first.

ChALkeR:
Going to merge this soon, if no objections. /cc @akreuzkamp again.
The conflicts are due to #304 being landed first, I will fix that directly before merging.

ChALkeR:
Btw, @pavelvasev, I expect rewriting some closures to prototype methods (as partially done here) to have a positive effect on the performance.

pavelvasev:
@ChALkeR thank you for info. Btw, when you write `somefuncname() { ..body.. }` instead of `function somefuncname() { ... body.. }` is it possible in modern js?..

ChALkeR:
@pavelvasev https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Classes. Note that it's not «real» classes, but syntax sugar around prototypes.

pavelvasev:
Thank you @ChALkeR !

ChALkeR:
The bad thing is that I had to name two methods `$onFocusChanged_` and `$onVisibleChanged_` for them to not override similar methods in `Item`. I will fix this later, for now I want to merge as-is to simplify further fixes.

PR #307, opened 2016-07-17
QColor cleanup

ChALkeR:
/cc @akreuzkamp @Plaristote @labsin 

The first commit is required to support initializing property types that require `new` usage (both «classes» and other prototype-based).

Creation of a `QColor` from an array was broken and not used.
Qt QML keeps the converted colors lower-case, testcase added.

`QColor` is not yet finished here, but it's already better than what we have now, and I think we should land this as is and continue improving it later.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -15,6 +15,7 @@ const path = require("path");
 
 const qtcoreSources = [
   "src/QmlWeb.js",
+  "src/polyfills/*.js",
   "src/qtbase/QObject.js",
   "src/qtbase/*.js",
   "src/modules/QtQml/Qt.js",
--- a/src/engine/QMLBoolean.js
+++ b/src/engine/QMLBoolean.js
@@ -1,0 +1,4 @@
+function QMLBoolean(val) {
+  return !!val;
+}
+QMLBoolean.plainType = true;
--- a/src/engine/QMLInteger.js
+++ b/src/engine/QMLInteger.js
@@ -1,3 +1,4 @@
 function QMLInteger(val) {
     return (val|0);
 }
+QMLInteger.plainType = true;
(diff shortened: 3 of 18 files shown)


ChALkeR:
Going to land this soon if no objections arise.

ChALkeR:
Added another commit which adds a magic method `.$get()` to QML types, if they implement it, then properties would return the result of a `.$get()` method instead of the actual property value. That is required for proper implementation of comparison for `color` type. Tests included.

Atm, the implementation could consume some extra memory for rapidly changing colors (it stores and instance of each occured color indefinetely), that could be easily fixed with a `WeakMap`, but we need a polyfill for that for older browsers.

ChALkeR:
Update: implemented a `WebMap`-based approach with a polyfill. Btw, we could use `WeakMaps` in other places for better memory management.

akreuzkamp:
I don't really like this, but I don't see a way around it (I don't like languages without operator overloads :D), so go for it :)
But are you sure, that the polyfill really works? It seems you're using objects as keys... If at all this is only supported by ECMA 6, isn't it?

ChALkeR:
@akreuzkamp No, it doesn't use objects as keys. The `$value` is a string representation of a color, so it's a map from strings to QColor instances. There is nothing `Weak` about that fallback, though — but that would only be noticable if one creates a lot of unique colors _and_ uses an old browser (FF < 6.0, IE < 11.0, Safari < 7.1, Opera < 23, Chrome < 36).

ChALkeR:
@akreuzkamp I added some comments to the polyfill file so we don't forget that =).

ChALkeR:
Argh. @akreuzkamp, sorry, I had to revert the WeakMap thing, it wasn't working with real WeakMaps — those can't have strings as keys. The only way to fix this would be to limit the number of created QColor instances before this falls back to the not-so-compatible impl, and allow the user to change that limit. I think something about 10000 would do (about 200-300 KiB).

ChALkeR:
Fix in #311.

PR #309, opened 2016-07-18
QmlWeb: split and move out of the engine dir

ChALkeR:
The namespace file now resides directly in `./src/` dir and is included first, immediately followed by the complete `qtcore` (as it doesn't depend on anything but the namespace).

Also moved the map of keyboard events to `keyboard.js`.

The only code change is that now `keyboardSignals` (the «map» of signal names) is being exported as `QmlWeb.keyboardSignals`, everything else is just code movement.

Refs: #143

/cc @akreuzkamp


ChALkeR:
Going to land this soon if no objections arise.

akreuzkamp:
lgtm

PR #311, opened 2016-07-20
QColor: introduce a limit to comparable colors

ChALkeR:
This adds a user-configurable limit of comparable colors to avoid the potential memory hit.

The limit is set to 10000 QColor instances, which should use about 200-300 KiB.

Refs: https://github.com/qmlweb/qmlweb/pull/307

/cc @akreuzkamp


Plaristote:
Looks good enough to me.

I'm wondering if we should also have a way to remove colors... but it sounds like implementing that automatically would be tedious, and giving a manual control over this might be pointless...

PR #312, opened 2016-07-20
Flow implicit size

stephenmdangelo:
Note, the check for `this.flow === undefined` is required because `Positioner` calls `Flow.layoutChildren` in its constructor, but the `flow` property and `Flow` enum aren't registered until later.  

`Flow` was already calling `layoutChildren` at the end of it's constructor, so this doesn't break anything, but allows us to use the `Flow` enum within `layoutChildren` (instead of the check `this.flow == 0`).

I don't know if there's a better way to protect against this in general.  Unless enums/properties are created BEFORE the call to `meta.super`, but I don't know if that breaks things in the general case.

In the analogous example in C++ where `layoutChildren` would be a pure virtual function,  the compiler warns you and you get a linking error: http://cpp.sh/5jrm. Not sure if this just means `Positioner` should just leave the responsibility of calling `layoutChildren` to it's child classes, or if this is a larger issue that can be solved generically for other classes as well?


stephenmdangelo:
Wait on this, I think the implicitHeight/Width may be incorrect if the opposite value is explicitly set.

stephenmdangelo:
Fixed now, added the two cases that weren't working to the test.

ChALkeR:
Does this affect only `Flow`? I remember seeing other similar code in the repo.

stephenmdangelo:
> Does this affect only Flow? I remember seeing other similar code in the repo (perhaps Grid?).

What bit are you referring to, the bit about calling `layoutChildren` in the `Positioner` constructor? `Grid.layoutChildren` would have the same problem, but avoids it by comparing to `0`, `1` instead of referencing `Grid.LeftToRight` or `Grid.TopToBottom`.

I believe the `implicitWidth` and `implicitHeight` of `Grid` are correct since it has explicit rows/columns, not the implicit ones in `Flow` based on it's width/height.

src/modules/QtQuick/Flow.js
```diff
@@ -55,9 +61,12 @@ registerQmlType({
             curVPos += child.height + this.spacing;
         }
     }
-    if (this.flow == 0)
-        this.implicitHeight = curVPos + rowSize;
-    else
-        this.implicitWidth = curHPos + rowSize;
+    if (this.flow === Flow.LeftToRight) {
```
ChALkeR: `this.Flow.LeftToRight`, let's not rely on globals.

src/modules/QtQuick/Flow.js
```diff
@@ -17,10 +17,16 @@ registerQmlType({
     this.flowChanged.connect(this, this.layoutChildren);
     this.layoutDirectionChanged.connect(this, this.layoutChildren);
     this.widthChanged.connect(this, this.layoutChildren);
+    this.heightChanged.connect(this, this.layoutChildren);
     this.layoutChildren();
   }
 
   layoutChildren() {
+    if (this.flow === undefined) {
+      /* Flow has not been fully initialized yet */
```
ChALkeR: Could you use `//` for single-line comments?

ChALkeR:
LGTM. This needs to be rebased, though.

src/modules/QtQuick/Flow.js
ChALkeR: See https://github.com/qmlweb/qmlweb/commit/032f73242906e6a19edff1189367ebbce52e6d30, for example.

stephenmdangelo:
Rebase, and changes applied.

ChALkeR:
Merged in f57d67a...6d9b2314, thanks!

PR #313, opened 2016-07-21
Property binding obey type

stephenmdangelo:


stephenmdangelo:
> Merging #313 into master will increase coverage by 0.01%

Well that's reason enough, right there!

ChALkeR:
Um. That codecov message wasn't supposed to be here.

stephenmdangelo:
> Um. That codecov message wasn't supposed to be here.

Also, this seems to be breaking "changed" signals from properly emitting for certain properties... looking into it...

stephenmdangelo:
So this didn't actually break anything, it just moved an existing bug on manually setting an enum to a string value into bindings.

e.g. the `Font` enum is full of strings, if you set `uppercase` to the `Font.capitalization` `enum` property, it gets converted to `NaN`. Previously this was only happening with `text.font.capitalization = Font.AllUppercase`, with this change it happens with `font.capitalization: Font.AllUppercase` as well.

So I'm also going to look into fixing the `Font` enum.  I remember some talk a while about about not wanting string enum values anyway.

ChALkeR:
@stephenmdangelo Yes, enums should be integers afaik, and we somewhy have string enums in some places. That is probably broken because enum helper converts those to integers.

The proper way of fixing that would be to check the actual values of the affected enums in Qt QML.

stephenmdangelo:
> The proper way of fixing that would be to check the actual values of the affected enums in Qt QML.

Yep, working on it. I've got Font fixed, but there are some more in Text alignment as well.

stephenmdangelo:
Font and Text horizontal alignment enums are now integers.

(Unfortunately I have to use `this.Text` since there is already a global `Text` function, so the enums aren't registered globally...)

stephenmdangelo:
This is rebased on master and should be good if someone would like to review.

stephenmdangelo:
Just removed an old reference to a function I had added, but then removed in Text.js.

src/modules/QtQuick/Font.js
```diff
@@ -7,24 +7,50 @@ registerQmlType({
...
+        case Font.Black:
+          return "900";
+        default:
+          return "normal";
+      }
+    }
+
     this.boldChanged.connect(this, this.$onBoldChanged);
         this.capitalizationChanged.connect(function(newVal) {
+            newVal = ["none", "uppercase", "lowercase", "small-caps", "capitalize"][newVal]
```
ChALkeR: This function will be cleaned up later, but it's a bad practice to override input this way.
This variable should have been differently named.
Don't bother, though, I will fix this later.

ChALkeR:
LGTM.
This needs a rebase and a tests run, though.

src/modules/QtQuick/Text.js
```diff
@@ -103,20 +101,26 @@ registerQmlType({
...
+        break;
+      case this.Text.AlignRight:
+        textAlign = "right";
+        break;
+      case this.Text.AlignHCenter:
+        textAlign = "center";
+        break;
+      case this.Text.AlignJustify:
+        textAlign = "justify";
+        break;
```
ChALkeR: This could be cleaned up a bit, will do that later in a separate commit.

src/modules/QtQuick/Font.js
stephenmdangelo: I think I've now made this function clearer. It was confusing before will all of the random conditional overwrites/uses of newVal.

stephenmdangelo:
Should be good now.

ChALkeR:
Thanks, landed in f098301...dd85704.

ChALkeR:
Note that `capitalizationToTextTransform` doesn't seem to be covered by tests. As well as `weightToCss(Font.Normal)`.

PR #318, opened 2016-07-31
Fixes issue with updating wrapping Text implicit height when width is changed

hhaveri:


ChALkeR:
Thanks! LGTM, landed with a commit message change.

Btw, you might want to add that email to your profile at https://github.com/settings/emails, so that your avatar and profile link would be showed next to the commit.

PR #319, opened 2016-08-01
Add Heikki Haveri to AUTHORS

ChALkeR:
@hhaveri, could you confirm that this is the preferred name spelling and email address?


hhaveri:
Looks ok to me

Issue #316, opened 2016-07-30
Properly resolve url type properties 


stephenmdangelo:
Currently, properties of type `url` are just set as strings. Instead, they should be resolved relative to the file I which they are set.

stephenmdangelo:
I'll push what I have so far in the next few days. I've created a QMLUrl plainType that calls resolvePath when set. There are just some context basePath things that I'm working out to make it all work in the same way as Qt QML.

stephenmdangelo:
Fixed by #320

PR #320, opened 2016-08-02
Qml url

stephenmdangelo:
This depends on #313 for a couple of the tests to succeed.  Without it, the result of non-simple bindings aren't even passed to the QMLUrl constructor.

I think the logic of `$rootBasePath` makes sense, but I'm not 100% sold on it. It currently makes any absolute URLs within the context of QML relative to the root QML item's path, which is not necessarily the same as the webpage's root URL. It makes it easier to move a qmlweb project with absolute URLs around within a webpage's directory structure, but it makes it impossible to reference URLs outside of the qmlweb sandbox, without a full `http://...`.

Maybe there shouldn't be a concept of `$rootBasePath` and it's just up to the webpage admin to setup URL symlinks and the like accordingly.

### DIFF
--- a/src/engine/QMLEngine.js
+++ b/src/engine/QMLEngine.js
@@ -478,7 +478,7 @@ class QMLEngine {
       return this.components[name];
     }
 
-    const file = `${this.$basePath}${name}.qml`;
+    const file = QmlWeb.engine.$resolvePath(`${name}.qml`);
     this.ensureFileIsLoadedInQrc(file);
     const tree = QmlWeb.convertToEngine(QmlWeb.qrc[file]);
     tree.$file = file;
--- a/src/engine/QMLEngine.js
+++ b/src/engine/QMLEngine.js
@@ -474,19 +474,33 @@ class QMLEngine {
     let tree;
     if (uri.scheme === "qrc://") {
       tree = QmlWeb.qrc[uri.path];
+      if (!tree) {
+        return undefined;
+      }
+      // QmlWeb.qrc contains pre-parsed Component objects, but they still need
+      // convertToEngine called on them.
+      tree = QmlWeb.convertToEngine(tree);
     } else {
-      const src = QmlWeb.getUrlContents(file);
+      const src = QmlWeb.getUrlContents(file, true);
       if (!src) {
-        console.log("Can not load file [", file, "]");
+        console.error("QMLEngine.loadComponent: Failed to load:", file);
         return undefined;
       }
 
-      QmlWeb.loadParser();
-      console.log("Loading file [", file, "]");
-      tree = QmlWeb.parse(src, QmlWeb.parse.QMLDocument);
+      console.log("QMLEngine.loadComponent: Loading file:", file);
+      tree = QmlWeb.parseQML(src, file);
     }
 
-    tree = QmlWeb.convertToEngine(tree);
+    if (!tree) {
+      return undefined;
+    }
+
+    if (tree.$children.length !== 1) {
+      console.error("QMLEngine.loadComponent: Failed to load:", file,
+        ": A QML component must only contain one root element!");
+      return undefined;
+    }
+
     tree.$file = file;
     this.components[file] = tree;
     return tree;
--- a/src/engine/QMLEngine.js
+++ b/src/engine/QMLEngine.js
@@ -337,8 +337,6 @@ class QMLEngine {
         name = name.substr(0, name.length - 1);
       }
     }
-    // TODO if nameIsDir, we have also to add `name` to importPathList() for
-    // current component...
 
     let content = this.qmldirsContents[name];
     // check if we have already loaded that qmldir file
--- a/src/engine/QMLEngine.js
+++ b/src/engine/QMLEngine.js
@@ -17,8 +17,11 @@ class QMLEngine {
     this.running = false;
     this.rootElement = element;
 
-    // List of available Components
+    // Cached component trees (post-QmlWeb.convertToEngine)
     this.components = {};
+
+    // Cached parsed JS files (post-QmlWeb.jsparse)
+    this.js = {};
 
     // List of Component.completed signals
     this.completedSignals = [];
@@ -93,22 +96,6 @@ class QMLEngine {
     }
   }
 
-  ensureFileIsLoadedInQrc(file) {
-    if (QmlWeb.qrc.hasOwnProperty(file)) {
-      return;
-    }
-
-    const src = QmlWeb.getUrlContents(file);
-    if (!src) {
-      console.log("Can not load file [", file, "]");
-      return;
-    }
-
-    QmlWeb.loadParser();
-    console.log("Loading file [", file, "]");
-    QmlWeb.qrc[file] = QmlWeb.parse(src, QmlWeb.parse.QMLDocument);
-  }
-
   // eslint-disable-next-line max-len
   /** from http://docs.closure-library.googlecode.com/git/local_closure_goog_uri_uri.js.source.html
    *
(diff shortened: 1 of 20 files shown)


stephenmdangelo:
This is to fix #316

stephenmdangelo:
Fully formed URLs (e.g. http://, file://, etc.) weren't working in the initial change. Now they are and I've added a test for them.

ChALkeR:
This should be rebased now, as #313 is merged.

src/engine/QMLUrl.js
```diff
@@ -0,0 +1,28 @@
+function QMLUrl(val) {
```
ChALkeR: Not sure if this is the correct location for the new `url` type. Perhaps put it in `qtbase`?

/cc @akreuzkamp, @labsin — thoughts?

src/engine/QMLUrl.js
```diff
@@ -0,0 +1,28 @@
...
+      }
+    }
+    resolvedPath = resolvedPath.slice(i);
+    resolvedPath.unshift("");
+    return resolvedPath.join("/");
+  }
+
+  return resolvedPath;
+}
+QMLUrl.plainType = true;
```
ChALkeR: Plain types return raw values, while normal types (e.g. #299) are classes with their own methods.
Does `url` instances have any methods or behaviour that distinguishes them from plain strings, apart from the initial resolving to the base path upon assigment?

ChALkeR:
Mostly LGTM, some questions about the new `url` type.

Could you please amend the commits here to include 
```

PR-URL: https://github.com/qmlweb/qmlweb/pull/320
Reviewed-By: Сковорода Никита Андреевич <chalkerx@gmail.com>
```
at the bottom of each commit?

src/engine/QMLUrl.js
stephenmdangelo: http://doc.qt.io/qt-5/qml-url.html

There doesn't SEEM to be any methods specified in the docs. I think it's basically just the initial resolve.

ChALkeR: Checked via enumerating the methods. There are `preserve` and `destroy` which seem to be non-standard, but I'm not sure if we want to reimpliment those.

Ok, makes sense to keep it this way for now, then.

ChALkeR: Argh, enumerating the methods does not work for native Qt objects. So I might have missed something.

src/engine/QMLEngine.js
```diff
@@ -493,12 +498,23 @@ class QMLEngine {
 
   // Return a path to load the file
   $resolvePath(file) {
+    if (this.$rootBasePath === undefined || this.$basePath === undefined) {
+      return file;
+    }
+
+    // "Root" references in QML should still be relative to the root QML file
+    if (file.indexOf("/") === 0 && this.$rootBasePath !== undefined) {
```
stephenmdangelo: Are you good with this concept of rootBasePath? It requires anything outside of wherever the root QML file is stored to have a full HTTP path. I've started playing around with the qrc system in qmlweb and feel like that might be a better approach to create an "insulated" file structure.

ChALkeR: > Are you good with this concept of rootBasePath?

I think yes, we could improve things later if needed, and this looks like a clear improvement to me atm.

I am not sure if we should store `$rootBasePath` though, as we could just do `const rootBasePath = this.$basePath.slice(0, this.$basePath.lastIndexOf('/'))` directly whenever it's used, and `this.$rootBasePath === undefined` is equivalent to `this.$basePath === undefined` in the current code.

> It requires anything outside of wherever the root QML file is stored to have a full HTTP path.

Could you provide an example where this would break?

stephenmdangelo: > I am not sure if we should store $rootBasePath though, as we could just do...

I don't think that will work.  `$rootBasePath` is only set when then initial QML file is loaded from the `data-qml` attribute of the `body` tag, whereas `$basePath` is set to the path of whatever QML component file is currently being evaluated (and, in this change, in a few more places to ensure that the correct `$basePath` is used for evaluating relative paths).

So, for example, if you have a directory structure like:

```
http://www.example.com
/index.html
/test/test.qml
/test/foo/Bar.qml
/test/foo/image.png
```

And `index.html` has:

```
<body data-qml="test/test.qml">
```

test.qml:

```
import "foo"

Bar {
}
```

Bar.qml

```
import QtQuick 2.0
Item {
    property url path: "image.png" // $basePath would be set to "test/foo" when this evaluated so that the path would resolve to "test/foo/image.png".  $rootBasepath would always be "test/".
}
```

> Could you provide an example where this would break?

In the above example, if the path in `Bar.qml` was `/image.png`, it will now resolve to `http://www.example.com/test/image.png`. Whereas `/image.png` in the normal HTML/CSS-case would resolve to `http://www.example.com/image.png`.

In the realm of file-based QML, `/image.png` would resolve to `file:///image.png`, which to me feels more like `http://www.example.com/image.png`. So I think it makes the most sense to leave out the `$rootBasePath` stuff.  I think what I was trying to accomplish is more of a job for the `qrc` system.

I feel the best approach going forward would be to expand upon the `qrc` system to behave more like QML (e.g. if a Component file is within the `qrc` file system, all paths are relative to the `qrc` file system unless fully resolved). And add some support for "storing" images in the `qrc` filesystem (probably just as an HTML-relative URL, maybe with an option for base64 encoding if that's something someone wants).

stephenmdangelo:
@ChALkeR Alright, so I think this mostly happy with this now.

It now resolves all URLs to an absolute URL (e.g. including http://...).  This is what Qt QML does (e.g. file URLs will print out "file:///path/to/file").  This is also to pave the way to be able to have "qrc://" URLs that reference things within the resource system (and to have any URLs resolved within files from the resource system to be qrc:// URLs (as explained here: http://doc.qt.io/qt-5/qml-url.html).

Downside: All QmlWeb.qrc stuff is broken :) (unless you go and put in the full http://... url as they key).

I am working on a proper solution to this, but would like to get this initial stuff reviewed while I do that.

My plan is that you should be able to have "qrc:///root.qml" as your "data-qml", this will then look for "/root.qml" in QmlWeb.qrc, "qrc:///" will then be the $basePath, and any subsequent relative URLs will be "qrc:///...". Any time we try to load a "qrc:///path" URL, we look for "/path" in QmlWeb.qrc.

stephenmdangelo:
To further elaborate, there are some places (like in Loader) that use QmlWeb.qrc as a cache (by calling `QMLEngine.ensureFileIsLoadedInQrc`).  This is incorrect and will be removed, as the resource system is not a cache (the parsed content of a Component can always still be cached for performance reasons).

stephenmdangelo:
Alright, I think this all makes sense now if anyone would like to review :)

ChALkeR:
@stephenmdangelo Thanks, I will take a look at this soon! =)

stephenmdangelo:
Another thing I think I'll add later is to have `file://` URLs resolve to the root of the website (or maybe this is where a rootPath could be user-configured). I think this would make it easier to share non-qrc-based projects between Qt QML and QmlWeb by not having to hard-code in the domain into absolute paths.

ChALkeR:
Re: qrc changes — does gulp-qmlweb have to be changed to support this?

src/engine/modules.js
```diff
@@ -250,23 +249,21 @@ function construct(meta) {
...
     QmlWeb.executionContext = meta.context;
 
-    const Qt = QmlWeb.Qt;
+    let filePath;
     if (qdirInfo) {
-      // We have that component in some qmldir, load it from qmldir's url
-      component = Qt.createComponent(`@${qdirInfo.url}`);
+      filePath = qdirInfo.url;
+    } else
+    if (classComponents.length === 2) {
```
ChALkeR: `else if` should be on one line.

ChALkeR:
Mostly LGTM, with a question and a nit =).

stephenmdangelo:
gulp-qmlweb does not require any changes, but the way that the QmlWeb.qrc variable is used has changed. So anyone currently using gulp-qmlweb will have broken code, unless they change their root file in the body data-qml to be "qrc:///root.qml". Maybe I should make a test case so people know how to use it. I'll be able to tidy this up next week.

ChALkeR:
@stephenmdangelo In fact, if you split this PR in two (first two commits separate from the remaining two commits), I would gladly merge the first one today, as those look good to me now. :wink:

stephenmdangelo:
@ChALkeR Since the QMLUrl commit changes urls to be absolute (e.g. including the http://), all QmlWeb.qrc lookups will fail anyway, so I think it makes the most sense that it all comes in together.  I'll work on a test case today.

stephenmdangelo:
Tests added (which found some issues :)) I think this is good now...

src/engine/modules.js
ChALkeR: Will fix when landing.

ChALkeR:
Landed in 988dbba...9005e64.

ChALkeR:
Thanks! =)

PR #321, opened 2016-08-05
QtQuick/Item: Only set "parent" property for Item-inherited items

stephenmdangelo:
With the change to giving Timer a read-only "parent" property, it was
no longer added to resources (and logged out exceptions).

### DIFF
--- a/src/modules/QtQuick/Item.js
+++ b/src/modules/QtQuick/Item.js
@@ -186,11 +186,10 @@ registerQmlType({
     this.$updateVGeometry(newParent, oldParent, propName);
   }
   $onDataChanged(newData) {
+    const QMLItem = QmlWeb.getConstructor("QtQuick", "2.0", "Item");
     for (const i in newData) {
       const child = newData[i];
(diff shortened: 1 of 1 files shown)


stephenmdangelo:
@ChALkeR This fixes an issue I introduced in #287

src/modules/QtQuick/Item.js
```diff
@@ -188,9 +188,7 @@ registerQmlType({
   $onDataChanged(newData) {
     for (const i in newData) {
       const child = newData[i];
-      if (child.hasOwnProperty("parent")) {
-        // Seems to be an Item.
-        // TODO: Use real inheritance and ask using instanceof.
+      if (child instanceof QmlWeb.getConstructor("QtQuick", "2.0", "Item")) {
```
ChALkeR: Could you save this as `QMLItem` at the top of the function (i.e. outside the loop)?

stephenmdangelo: Done :)

ChALkeR:
Thanks!

PR #322, opened 2016-08-06
MouseArea: fix hover & clicked events

hhaveri:
Parts of MouseArea got broken in commit 9b1d2bbee6993fd29d4ebf46ac4be2a43695d65c


ChALkeR:
Yes, thanks for noticing that!

PR #323, opened 2016-08-06
Loader: fix callOnCompleted method

hhaveri:
Parts of Loader got broken in commit 6f746cb5656e942ec94397fcf1b22df27354a8ca. Having not much competence in javascript / qmlweb, just copied related implementation from Repeater that seems to work better - at least for now.


ChALkeR:
LGTM, thanks!

PR #325, opened 2016-08-08
(WIP) tests: Test release outside of MouseArea

stephenmdangelo:
I can't find a way to access PhantomJS's sendEvent call, except via a
Render test.  At this point it can send the mouse events, but they aren't
making it to the MouseArea.


stephenmdangelo:
Currently failing tests for #256.

ChALkeR:
Thanks!

ChALkeR:
I will take a look at those a bit later.

ChALkeR:
Looks like the dispatched event only goes to the top-level Karma window and is not invoked on the actual runner iframe (Karma uses those to run tests). Still not sure why is that.

In the meantime, you could just use browser API to test things:
```js
    var event = new MouseEvent(type, { 'bubbles': true, 'cancelable': true });
    mouse_area.dom.dispatchEvent(event);
```

ChALkeR:
Uh-oh, `new MouseEvent` seems to be not supported by PhantomJS (it uses an old WebKit version).
Do this instead:
```qml
import QtQuick 2.0

Rectangle {
  id: qml
  width: 200
  height: 200
  property bool pressed: mouse_area.pressed
  MouseArea {
    id: mouse_area
    width: 100
    height: 100

    function sendEevent(type, x, y, button) {
      var event = document.createEvent('MouseEvents');
      event.initMouseEvent(type, true, true, window, 0, 0, 0, x, y,
                           false, false, false, false, button || 0, null);
      /*
      var event = new MouseEvent(type, {
        bubbles: true, cancelable: true,
        clientX: x, clientY: y,
        button: button
      });
      */
      mouse_area.dom.dispatchEvent(event);
    }
  }

  Timer {
    interval: 100
    running: window !== undefined
    onTriggered: {
      mouse_area.sendEevent("mousedown", 0, 0);
      console.log(qml.pressed);
      mouse_area.sendEevent("mouseup", 0, 0);
      console.log(qml.pressed);
      mouse_area.sendEevent("mousedown", 0, 0);
      console.log(qml.pressed);
      mouse_area.sendEevent("mouseup", 11, 11);
      console.log(qml.pressed);
    }
  }
}
```

ChALkeR:
The proposed approach has the drawback is that it doesn't check that the mouse event is in fact catched by the correct element, instead we just manually emit mouse events on elements. But that's still better than nothing =).

Btw, you could manually dispatch the last `mouseup` event on the `document.body` instead.

tests/phantom.callback.js
```diff
@@ -1,14 +1,21 @@
 /* global page */
+/*eslint prefer-spread: 1 */
```
ChALkeR: Note: 4e36da24a8a20009b351a982972e3601d3d3e177 disabled `prefer-spread` rule in `tests` dir (as it should have been done from the start), so this line isn't needed any more.

ChALkeR:
Ok, I cleaned up my approach and pushed it in 240fc335765a4e4bc5eaf111bd46619af2e4ad8f.

Thanks again!

PR #326, opened 2016-08-15
QMLEngine: simplify import path related code style

ChALkeR:
No actual changes here, code style only.

IMO, it looks more readable this way.

/cc @stephenmdangelo


stephenmdangelo:
LGTM

ChALkeR:
Thanks, landed.

PR #329, opened 2016-08-21
Implement unique connections, reduce code complexity

ChALkeR:
This adds `Signal.UniqueConnection` as a possible last argument to `signal.connect(…)` (see http://doc.qt.io/qt-5/qt.html#ConnectionType-enum).

They are not present on the `QmlWeb.Qt` object, but rather on `QmlWeb.Signal` instead, but that is probably fine as `Qt` is supposed to be qml-specific.

`AutoConnection` (default) and `UniqueConnection` are supported.

This also reduces the cyclomatic complexity of the code in `./src/` from 40 to 35. 35 is reached at `$updateVGeometry` and `$updateHGeometry`, I am working on reducing that furhter by cleaning up the code (there are many unneeded things there), but I would prefer this to land first.

Note: `QmlWeb.$ease` is currently whitelisted in code complexity check, as it has cyclomatic complexity of 78 — that's the method that implements all the easing functions. We could split that into separate easing functions, though.

/cc @akreuzkamp


src/qtbase/Signal.js
```diff
@@ -24,6 +24,15 @@ class Signal {
     QmlWeb.QMLProperty.popEvalStack();
   }
   connect(...args) {
+    let type = Signal.AutoConnection;
+    if (typeof args[args.length - 1] === "number") {
+      type = args.pop();
+    }
+    if (type & Signal.UniqueConnection) {
+      if (this.isConnected(...args)) {
```
ChALkeR: Perhaps should join this into a single `if`, no sense in nesting here.

akreuzkamp:
ship it :)

ChALkeR:
@akreuzkamp Thanks! =)

src/qtbase/Signal.js
ChALkeR: On the second thought, it looks more clear this way.

PR #331, opened 2016-08-21
qml: don't override default constructors in loadImports()

ChALkeR:
I am currently reviewing and cleaning up the last file left — `qml.js`, and that process was stuck on that line.

This PR is only about the actual change, one we figure out what should be done there I will clean up the code accordingly, perhaps even removing some unneeded data structures. This is most likely wrong, I am not sure what's going on here, this is why I am asking =).

What `loadImports()` does now is that it both overrides the default `constructors` _and_ saves those in `perImportContextConstructors` for further use in `construct`. Perhaps that's a mistake?

`loadImports` resets the scoped (global-ish) `constructors` atm on each `loadImports()` call.

Note that `registerGlobalQmlType()` (and `registerQmlType` with `global` option) add new elements to `constructors` — those are not removed by `loadImports()`, though, that's probably as expected.

The complete list of places where that `constructors` object is used:
- https://github.com/qmlweb/qmlweb/blob/master/src/engine/QMLEngine.js#L58-L63
- https://github.com/qmlweb/qmlweb/blob/master/src/engine/QMLProperty.js#L23-L56
- https://github.com/qmlweb/qmlweb/blob/master/src/modules/QtQml/Component.js#L14

So this is basically a question of what exactly should be used in those places? Perhaps those lines should be changed to include something else than this `constructors` object?

Note: I could be missing something, and it might in fact be as designed.

/cc @akreuzkamp @stephenmdangelo @pavelvasev

### DIFF
--- a/src/engine/qml.js
+++ b/src/engine/qml.js
@@ -142,7 +142,7 @@ const perImportContextConstructors = {};
 
 function loadImports(self, imports) {
   const mergeObjects = QmlWeb.helpers.mergeObjects;
-  constructors = mergeObjects(modules.Main);
+  let constructors = mergeObjects(modules.Main);
   if (imports.filter(row => row[1] === 'QtQml').length === 0 &&
       imports.filter(row => row[1] === 'QtQuick').length === 1) {
     imports.push(['qmlimport', 'QtQml', 2, '', true]);


stephenmdangelo:
+1 I only left the `constructors` stuff as-is in `loadImports` as I didn't want to possibly break something. This was back when I was a naive youngin and the qmlweb code-base mystified me.

ChALkeR:
@stephenmdangelo I am still not sure what to with the list above, where `constructors` object is used directly.

 1. [QMLEngine.js#L58-L63](https://github.com/qmlweb/qmlweb/blob/master/src/engine/QMLEngine.js#L58-L63) — was introduced in 098962f98451f3bf3459c7a766821f7e8955aa0f, @akreuzkamp. I am not even sure what that code does. @akreuzkamp, could you explain?
 2. [QMLProperty.js#L23-L56](https://github.com/qmlweb/qmlweb/blob/master/src/engine/QMLProperty.js#L23-L56) uses `constructors` to instantiate property values. Perhaps that should use `perImportContextConstructors`? Not sure. @stephenmdangelo, could you take a look?
 3. [Component.js#L14](https://github.com/qmlweb/qmlweb/blob/master/src/modules/QtQml/Component.js#L14) — same as the previous, perhaps it should use `perImportContextConstructors` of the current file?

The issue with `constructors` is that it's a single entity for all files, while that should be actually scoped per context. Only three places are affected, and those should be somehow fixed.

ChALkeR:
Note: `constructors` have been renamed to `QmlWeb.constructors`, but I'm still not sure if we should keep those (same reasons as above).

Plaristote:
@ChALkeR:
About QMLEngine.js#L66-73 from 098962f, it's currently only used to access `this.Component`.
If we assume Component is the only type of object that you inherit from and that can be accessed that way, then that's all this code does, at least as far as I can tell. I'm not sure whether this assertion is true or not.

Plaristote:
It seems like this change has already made its way into the code-base (see `src/engine/modules.js:129` as of bcd5b08).

PR #338, opened 2016-09-15
QMLBinding: add a simpler variables resolver

### LINKED ISSUE #337: `this` should be bound to the current context
Testcase:
```qml
import QtQuick 2.0

Rectangle {
  color: "yellow"
  width: 200
  height: this.width
}
```

ChALkeR:
~~Also fixes `this` support in bindings, so this fixes #337.~~ — that part landed as 45aee57.

`In progress` because this ~~needs tests~~ and perhaps benchmarks (I am not even sure if this is actually faster at this point), everything else is ready here.

@stephenmdangelo, what do you think?

### DIFF
--- a/src/engine/QMLBinding.js
+++ b/src/engine/QMLBinding.js
@@ -26,8 +26,37 @@ class QMLBinding {
   }
 
   eval(object, context, basePath) {
-    // .call is needed for `this` support
-    return this.impl.call(object, object, context, basePath);
+    if (this.impl) {
+      // .call is needed for `this` support
+      return this.impl.call(object, object, context, basePath);
+    }
+    if (this.src === "this") {
+      return object;
+    }
+
+    const path = this.path;
+    const varname = path[0];
+
+    QmlWeb.executionContext = context;
+    if (basePath) {
+      QmlWeb.engine.$basePath = basePath;
+    }
+
+    let base;
+    if (varname in object) {
+      base = object[varname];
+    } else if (varname in context) {
+      base = context[varname];
+    } else if (varname in QmlWeb) {
+      base = QmlWeb[varname];
+    } else {
+      throw new ReferenceError(`${varname} is not defined`);
+    }
+
+    for (let i = 1; i < path.length; i++) {
+      base = base[path[i]];
+    }
+    return base;
   }
 
 /**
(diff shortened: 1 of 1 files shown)


src/engine/QMLBinding.js
```diff
@@ -24,19 +25,68 @@ class QMLBinding {
     };
   }
 
+  eval(object, context, basePath) {
+    if (this.impl) {
+      // This is needed for `this` support
```
ChALkeR: perhaps need to better explain that this comment is related to `.call()` instead of direct `.impl(…)`.

src/engine/QMLBinding.js
```diff
@@ -24,19 +25,68 @@ class QMLBinding {
...
+      return this.impl.call(object, object, context, basePath);
+    }
```
ChALkeR: Perhaps add an extra newline here.

src/engine/QMLBinding.js
stephenmdangelo: Yeah, that confused me at first. Maybe: ".call is required for `this` support"

stephenmdangelo:
Assuming this is faster, I'm good with it.  If it's not any faster, it just seems like added complexity.  I assume you have some theory as to why this would be faster?

ChALkeR:
@stephenmdangelo It wouldn't need to create `Function` wrappers for simple variables.

Also, a significant part of this is needed for `this` support. I will split this into two commits, perhaps.

ChALkeR:
Split into two commits. Tests and benchmarks are still missing =).

src/engine/QMLBinding.js
```diff
@@ -26,16 +26,54 @@ class QMLBinding {
...
+    if (this.impl) {
+      // .call is needed for `this` support
+      return this.impl.call(object, object, context, basePath);
+    }
+    if (this.src === "this") {
+      return object;
+    }
+
+    const path = this.path;
+    const varname = path[0];
```
ChALkeR: what if `path[0]` is `this`? I should check that.

ChALkeR:
I added tests to the `this` fix and landed that as 45aee5730edd508c7d36639a0e3ce6312780a740.

This one is rebased and contains only a simplier variables resolver now.

ChALkeR:
Going to close this for now, perhaps will return to this later.

PR #339, opened 2016-09-15
QMLEngine: Fix loadFile to work with subdirectories

hhaveri:
Should fix the problem of autoloading for example "./qml/index.qml" resulting in trying to load ./qml/qml/index.qml" instead.

### DIFF
--- a/src/engine/QMLEngine.js
+++ b/src/engine/QMLEngine.js
@@ -142,7 +142,7 @@ class QMLEngine {
   }
 
   extractFileName(file) {
-    return file.split('\\').pop().split('/').pop();
+    return file.split(/[\/\\]/).pop();
   }
 
   // Load file, parse and construct (.qml or .qml.js)
--- a/src/engine/QMLEngine.js
+++ b/src/engine/QMLEngine.js
@@ -141,6 +141,10 @@ class QMLEngine {
     return basePath.join("/");
   }
 
+  extractFileName(file) {
+    return file.split('\\').pop().split('/').pop();
+  }
+
   // Load file, parse and construct (.qml or .qml.js)
   loadFile(file, parentComponent = null) {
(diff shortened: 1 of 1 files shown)


ChALkeR:
/cc @stephenmdangelo 

As of lint failures — `quotes` and `max-len` issues should be fixed, the latter could be done with an intermediate variable, `this.loadComponent(this.$resolvePath(this.extractFileName(file)));` is just too complex. Not sure about `newline-per-chained-call` — perhaps it would make sense to disable that rule.

src/engine/QMLEngine.js
```diff
@@ -141,6 +141,10 @@ class QMLEngine {
     return basePath.join("/");
   }
 
+  extractFileName(file) {
+    return file.split('\\').pop().split('/').pop();
```
ChALkeR: Could also just use `split(/[\/\\]/)`.

hhaveri: Not sure why, but that does not seem to work for me. Basically makes it try to load "./qml/.,qml,index.qml"

ChALkeR: @hhaveri Well, I meant a single `split` istead of two ones. `pop()` is still needed here =).

hhaveri: Yes, just noticed the missing pop() myself also.

ChALkeR:
About `newline-per-chained-call` — in the current configuration, it allows having chains of 2 calls on a single line. So everything is good lint-side.

Actual changes LGTM, but I would prefer a comment from @stephenmdangelo before landing this.

Thanks!

stephenmdangelo:
Sorry, not ignoring this, just haven't had a chance to look at this today. I'll review it tomorrow.

ChALkeR:
Landed in b7ecb5e, thanks!

PR #340, opened 2016-09-20
engine/properties: Set engine $basePath back after evaluating a slot

stephenmdangelo:
Slots can be invoked as the result of a property changing, and they
shouldn't leave the engine $basePath in a bad state.

### DIFF
--- a/src/engine/properties.js
+++ b/src/engine/properties.js
@@ -226,11 +226,16 @@ function connectSignal(item, signalName, value, objectScope, componentScope) {
     for (const j in item[signalName].parameters) {
       params.push(item[signalName].parameters[j].name);
     }
+    // Wrap value.src in IIFE in case it includes a "return"
     value.src = `(
       function(${params.join(", ")}) {
         QmlWeb.executionContext = __executionContext;
+        QmlWeb.engine.$oldBasePath = QmlWeb.engine.$basePath;
         QmlWeb.engine.$basePath = "${componentScope.$basePath}";
-        ${value.src}
+        (function() {
+          ${value.src}
+        })();
+        QmlWeb.engine.$basePath = QmlWeb.engine.$oldBasePath;
       }
     )`;
     value.isFunction = false;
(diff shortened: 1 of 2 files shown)


src/engine/properties.js
```diff
@@ -226,11 +226,16 @@ function connectSignal(item, signalName, value, objectScope, componentScope) {
     for (const j in item[signalName].parameters) {
       params.push(item[signalName].parameters[j].name);
     }
+    // Wrap value.src in IIFE in case it includes a "return"
     value.src = `(
       function(${params.join(", ")}) {
         QmlWeb.executionContext = __executionContext;
+        var bp = QmlWeb.engine.$basePath;
```
ChALkeR: Probably another name is needed in case if there is a `bp` property.
E.g. `__basePath` could do.

src/engine/properties.js
```diff
@@ -226,11 +226,16 @@ function connectSignal(item, signalName, value, objectScope, componentScope) {
     for (const j in item[signalName].parameters) {
       params.push(item[signalName].parameters[j].name);
     }
+    // Wrap value.src in IIFE in case it includes a "return"
```
ChALkeR: Perhaps add a testcase for that?

src/engine/properties.js
stephenmdangelo: I decided to hide it with QmlWeb.engine, figure that's the safest bet.

src/engine/properties.js
stephenmdangelo: Done.

src/engine/properties.js
ChALkeR: Ah, looks good.

ChALkeR:
LGTM, merging.

stephenmdangelo:
This actually needs ANOTHER fix. I have to add a "try...finally" around it to ensure that a slow throwing an exception doesn't mess up the basePath.

PR #345, opened 2016-10-12
modules/Text: Properly initialize wrapMode

stephenmdangelo:
Call $onWrapModeChanged to initialize CSS properties for the default
wrapMode of NoWrap.

Also removed double connections to text, wrapMode changed.

### DIFF
--- a/src/modules/QtQuick/Text.js
+++ b/src/modules/QtQuick/Text.js
@@ -49,9 +49,7 @@ QmlWeb.registerQmlType({
     this.color = "black";
     this.text = "";
 
-    this.textChanged.connect(this, this.$onTextChanged);
     this.widthChanged.connect(this, this.$onWidthChanged);
-    this.wrapModeChanged.connect(this, this.$onWrapModeChanged);
 
     this.font.boldChanged.connect(this, this.$onFontChanged);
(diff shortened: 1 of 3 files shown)


src/modules/QtQuick/Text.js
```diff
@@ -127,6 +125,7 @@ QmlWeb.registerQmlType({
   }
   Component$onCompleted() {
     this.$updateImplicit();
+    this.$onWrapModeChanged(this.wrapMode);
```
ChALkeR: Note: this now calls `$updateImplicit()` two times. Perhaps something could be done with that?
If the `$onWrapModeChanged` always calls `$updateImplicit`, perhaps the first line could be removed (replaced with a comment that `$updateImplicit()` needs to be called, but that it's already called by `$onWrapModeChanged`).

ChALkeR:
I decided to land this and fix calling `$updateImplicit` twice in a follow up commit.

Landed, thanks!

stephenmdangelo:
Thanks for merging all of these branches and doing the cleanup!

ChALkeR:
Cleanup landed in a59de4fb2d27021acc431cf2aa849c7384409721.

PR #347, opened 2016-10-24
Disconnect all signals on delete

stephenmdangelo:
I discovered this when the delegates of a Repeater are parented to a Column, as Positioner connects to various signals of it's children, and the Positioner is not the QObject parent of the delegates, but this could happen any time a signal on QObjectA is connected to a slot on QObjectB if QObjectB is not the parent of QObjectA.

I am open to suggestions on this fix. It seems overkill to add `$signals` as another list of things to cleanup (maybe `$tidyupList` could contain Objects with some extra information rather than just relying on the item type).

Also, there are some Signals that don't have an associated `obj` (like all of the `Keys.*` Signals). I'm going to test some more stuff with that, but I feel like a manual `item.Keys.pressed.connect(...)` could also result in a similar leak.

And the test is pretty internal and contrived, but I have confirmed it with the Chrome memory profiler.

### DIFF
--- a/src/qtbase/Signal.js
+++ b/src/qtbase/Signal.js
@@ -58,12 +58,14 @@ class Signal {
     // type meaning:
     //  1 = function, 2 = string
     //  3 = object with string method,  4 = object with function
+    // No args means disconnect everything connected to this signal
     const callType = args.length === 1
       ? args[0] instanceof Function ? 1 : 2
       : typeof args[1] === "string" || args[1] instanceof String ? 3 : 4;
     for (let i = 0; i < this.connectedSlots.length; i++) {
       const { slot, thisObj } = this.connectedSlots[i];
       if (
+        args.length === 0 ||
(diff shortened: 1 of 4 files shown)


src/qtbase/Signal.js
```diff
@@ -9,6 +9,11 @@ class Signal {
     this.signal.connect = this.connect.bind(this);
     this.signal.disconnect = this.disconnect.bind(this);
     this.signal.isConnected = this.isConnected.bind(this);
+
+    // XXX Fix Keys that don't have an obj for the signal
```
ChALkeR: Is this a `TODO`?

ChALkeR:
Landed in 02f995127d7b9d4992948e4bd961287aece37fdd...8f7fa3c632359f7d8d658c8f311b601984a476e1 with `s/XXX/TODO/`, thanks!

PR #348, opened 2016-10-25
adding QtQuick Binding

machinekoder:
Adds the QtQuick `Binding` component.


stephenmdangelo:
Are you able to add some tests?

machinekoder:
@stephenmdangelo Here they are.

ChALkeR:
Ignore the failing tests on Node.js 7.0.0, it's old `graceful-fs` fault. I will clear the build cache on Travis and re-start the test.

stephenmdangelo:
I think that `Binding` belongs in `modules/QtQml` instead of `modules/QtQuick`: http://doc.qt.io/qt-5/qtqml-qmlmodule.html

machinekoder:
@stephenmdangelo Yes, this is true.

src/modules/QtQml/Binding.js
```diff
@@ -0,0 +1,43 @@
+QmlWeb.registerQmlType({
+  module: "QtQuick",
```
stephenmdangelo: This should be `QtQml` as well.

stephenmdangelo:
I now see that my implementation breaks the `BindingBroken` test case... and adding a check that `this.value !== default_value` just means that whatever default value of `value` we use will end up not being able to be used within QML...

@ChALkeR Is there a way to actually provide a custom `setter` function for the `value` property, so we can know if it's been explicitly set or not?

stephenmdangelo:
> @ChALkeR Is there a way to actually provide a custom setter function for the value property, so we can know if it's been explicitly set or not?

I tried adding a `write` member to the property description and handle it like so in `QmlWeb.createProperty`:

```
  if (options.write) {
    const write = options.write;
    setter = function(newVal) {
      obj[write](newVal);
    };
  } else if (options.readOnly) {
...
```

But that only affect explicit sets of the value from JavaScript, not Bindings... we would probably have to put the concept right where `this.val = ` is inside of `QMLProperty.$setVal`.

stephenmdangelo:
@machinekoder Sorry for all of the noise, but I would suggest this implementation, with maybe a known-failing test if someone has a `Binding` with `value: undefined`. Note that the default value of `value` is now `undefined`, since it is that in Qt QML (at least in the version I tested with :)).

```
QmlWeb.registerQmlType({
  module: "QtQuick",
  name: "Binding",
  versions: /.*/,
  baseClass: "QtQml.QtObject",
  properties: {
    target: { type: "QtObject", initialValue: null },
    property: { type: "string", initialValue: "" },
    value: { type: "var", initialValue: undefined },
    when: { type: "bool", initialValue: true }
  }
}, class {
  constructor(meta) {
    QmlWeb.callSuper(this, meta);

    this.valueChanged.connect(this, this.$updateBinding);
    this.targetChanged.connect(this, this.$updateBinding);
    this.propertyChanged.connect(this, this.$updateBinding);
    this.whenChanged.connect(this, this.$updateBinding);
  }

  $updateBinding() {
    if (!this.when || !this.target
      || !this.target.hasOwnProperty(this.property)
      || this.value === undefined)
    {
      return;
    }
    this.target[this.property] = this.value;
  }
});
```

stephenmdangelo:
Or maybe keep `$property`, since it provides an optimization since it's more likely that `value` changes than `when` and `target`...

I think the key point I'm trying to make it to use `QMLProperty.set` instead of setting `QMLProperty.val` directly :)

machinekoder:
@stephenmdangelo Did you find time to look at the new patch?

tests/QtQml/Binding.js
```diff
@@ -0,0 +1,53 @@
...
+    var qml = load("Broken", this.div);
+    expect(qml.targetValue).toBe(0);
+    qml.when = true;
+    expect(qml.targetValue).toBe(0); // should not override anything
+  });
+
+  it("binding undefined var", function() {
+    var qml = load("Undefined", this.div);
+    qml.sourceValue = undefined;
+    expect(qml.targetValue).toBe(null); // should not work
```
stephenmdangelo: `qml.targetValue` SHOULD be `undefined` in this case (e.g. it is this way in Qt QML). So you should have: `expect(qml.targetValue).toBe(undefined);`, which will fail, but then you can add this test to `tests/failingTests.js` so we know it's something to fix in the future.

stephenmdangelo:
@ChALkeR Other than my last comment regarding a properly failing test, I'm good with this. When that's fixed, am I good to "Rebase and merge" right from the github UI here?

ChALkeR:
@stephenmdangelo Yes, just add the `Reviewed-By` (as in [AUTHORS](https://github.com/qmlweb/qmlweb/blob/master/AUTHORS#L18) and `PR-URL` (the full url, `https://github.com/qmlweb/qmlweb/pull/348`, not just `#348`, though GitHub will display it as such),  lines, please =).

That could be done after pressing the «Squash and merge» button.

machinekoder:
@stephenmdangelo I have updated the patch.

stephenmdangelo:
Manually merged to keep the update to qml-parser in a separate commit: c0fbbfb 3c09d8f

stephenmdangelo:
@machinekoder Thanks!

PR #349, opened 2016-10-26
Repeater delegate parent fix

stephenmdangelo:


src/modules/QtQuick/Repeater.js
```diff
@@ -71,6 +71,9 @@ QmlWeb.registerQmlType({
       if (this.$items.length > model) {
         // have more than we need
         this.$removeChildren(model, this.$items.length);
+        // Normally this is done in $insertChildren, but that won't be called
+        // in this case
+        this.count = this.$items.length;
```
ChALkeR: It looks confusing that `$insertChildren` and `$removeChildren` behave differently here.
As it doesn't seem viable to add this to `$removeChildren` (because `$removeChildren`+`$insertChildren` shouldn't update it twice), it should be removed from `$insertChildren`.

Then, the comment above won't be needed.

ChALkeR:
Landed in cf84a9eaaece88f74326c4d8cfaf4b22994fe38e...4d4ed1ca32b31de5dd4555a8477218812305c5f3, thanks!

Also I added 383e66e9a57b8e8a941599694cd82c708fc75987 to fix the `$insertChildren/$removeChildren` confusion.

PR #353, opened 2016-10-26
Loader empty source

stephenmdangelo:


src/modules/QtQuick/Loader.js
```diff
@@ -44,32 +44,41 @@ QmlWeb.registerQmlType({
...
-    if (newItem instanceof QMLComponent) {
-      qmlComponent = newItem.$createObject(this, {}, this);
-    }
-    qmlComponent.parent = this;
-    this.item = qmlComponent;
-    this.$updateGeometry();
-    if (this.item) {
-      this.loaded();
+
+    if (newItem) {
```
ChALkeR: Moving this one level deeper is not needed, just use this check instead:
```js
if (!newItem) {
  this.item = null;
  return;
}
```

src/modules/QtQuick/Loader.js
```diff
@@ -44,32 +44,41 @@ QmlWeb.registerQmlType({
...
-    const meta = { object: tree, context: this, parent: this };
-    const qmlComponent = new QMLComponent(meta);
-    qmlComponent.$basePath = QmlWeb.engine.extractBasePath(tree.$file);
-    qmlComponent.$imports = tree.$imports;
-    qmlComponent.$file = tree.$file;
-    QmlWeb.engine.loadImports(tree.$imports, qmlComponent.$basePath,
-      qmlComponent.importContextId);
-    const loadedComponent = this.$createComponentObject(qmlComponent, this);
-    this.sourceComponent = loadedComponent;
+    if (fileName) {
```
ChALkeR: Moving this one level deeper is not needed, just use this check instead:
```js
if (!fileName) {
  this.sourceComponent = null;
  this.$sourceUrl = fileName;
  return;
}
```

Not sure about `$sourceUrl`, perhaps it should be normalized to some falsy value. Or not.

ChALkeR:
Landed manually in 694ecdc...8c64ab1, thanks!

I applied minor changes to the if-then blocks (as mentioned above) before landing.

PR #354, opened 2016-11-02
QtQuick/MouseArea: set mouseX and mouseY before emitting positionChanged

stephenmdangelo:
This is what Qt QML does.

### DIFF
--- a/src/modules/QtQuick/MouseArea.js
+++ b/src/modules/QtQuick/MouseArea.js
@@ -61,9 +61,9 @@ QmlWeb.registerQmlType({
     this.dom.addEventListener("mousemove", e => {
       if (!this.enabled || !this.hoverEnabled && !this.pressed) return;
       const mouse = this.$eventToMouse(e);
-      this.positionChanged(mouse);
       this.mouseX = mouse.x;
       this.mouseY = mouse.y;
+      this.positionChanged(mouse);
     });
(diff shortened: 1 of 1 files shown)


stephenmdangelo:
This is a quick test that is different in Qt QML and qmlweb:

```
import QtQuick 2.0

Item {
    MouseArea {
        anchors.fill: parent
        onPositionChanged: {
            console.log(mouse.x, mouse.y, mouseX, mouseY)
        }
    }
}
```

stephenmdangelo:
@ChALkeR Travis seems to be complaining on un-related changes, but only for the first two versions of node?

ChALkeR:
@stephenmdangelo Looks like an eslint update, and it now finds more lint issues. To reproduce: `rm -rf node_modules/eslint*; npm i; npm run lint`.

This is not the fault of this PR, master is failing on lint.

The third Node.js version passes on CI only because it has an older version of eslint cached.

ChALkeR:
@stephenmdangelo Fixed in `master` in 675ec530d3, please rebase =).

stephenmdangelo:
@ChALkeR  Thanks. Rebased.

ChALkeR:
Landed in 02f995127d7b9d4992948e4bd961287aece37fdd, thanks!

PR #356, opened 2016-11-04
QtQuick/MouseArea: Emit positionChanged when pressed & cursor moves outside

stephenmdangelo:
The "mousemove" DOM event is only sent to the MouseArea's DOM element while
the cursor remains over the DOM element. This now attaches a temporary
event listener to "document" and translates the JavaScript MouseEvent
"clientX" and "clientY" properties to the MouseArea's coordinate space
(In regards to CSS transformations, this only takes into account translate
and scale transformations).

### DIFF
--- a/src/modules/QtQuick/MouseArea.js
+++ b/src/modules/QtQuick/MouseArea.js
@@ -36,19 +36,36 @@ QmlWeb.registerQmlType({
 
     this.dom.addEventListener("click", e => this.$handleClick(e));
     this.dom.addEventListener("contextmenu", e => this.$handleClick(e));
+    const handleMouseMove = e => {
+      if (!this.enabled || !this.hoverEnabled && !this.pressed) return;
+      this.$handlePositionChanged(e);
+    };
     const handleMouseUp = () => {
       this.pressed = false;
       this.pressedButtons = 0;
       document.removeEventListener("mouseup", handleMouseUp);
+      this.$clientTransform = undefined;
+      document.removeEventListener("mousemove", handleMouseMove);
     };
     this.dom.addEventListener("mousedown", e => {
       if (!this.enabled) return;
+      // Handle scale and translate transformations
+      const boundingRect = this.dom.getBoundingClientRect();
+      this.$clientTransform = {
+        x: boundingRect.left,
+        y: boundingRect.top,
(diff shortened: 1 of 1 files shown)


stephenmdangelo:
So this is a little hacky as we have to handle translate/scale transforms manually, but I couldn't find any other cross-browser way to make this work. I've tested this in Safari, Chrome and Firefox.

Firefox has `Element.setCapture` so that even outside of the `MouseArea` DOM element you get `e.offsetX/Y` relative to the `MouseArea` dom element, but Chrome and Safari don't have this.

Safari has `webkitConvertPointFromPageToNode` (https://developer.apple.com/reference/webkitjs/domwindow/1631559-webkitconvertpointfrompagetonode), which would allow us to easily convert `e.pageX/Y` to the `MouseArea` coordinate space (Chrome used to have this, but they removed it in blink with the explanation that `getBoundingClientRect` can be used... except it can't if you have a CSS rotation transformation... there's a hack to position a 1x1 child element at the relative point you want to map and call `getBoundingClientRect`, but that does a local->page mapping, not page->local mapping...).

I briefly looked into positioning 4 1x1 child elements in the corners of the `MouseArea` and mapping each of those and somehow deriving the inverse transformation matrix from that (or going all the way up the DOM tree and combining the parent CSS transform matrices), but that seems like overkill...

End rant. If anyone read this, thank you.

stephenmdangelo:
So I have to investigate a little further, but there may be an issue in Safari.

ChALkeR:
> So I have to investigate a little further, but there may be an issue in Safari.

I added a `work in progress` label for the time being, feel free to remove it when you decide to.

ChALkeR:
@stephenmdangelo This commit looks good to me in its current state. What is the issue in Safari that you were mentioning?

stephenmdangelo:
A colleague of mine was testing it out on a retina MacBook and it seemed like the scale was off (e.g. Stuff wasn't actually following the cursor) I've yet to look into it. We can always merge and fix later? Maybe merging will get some user out there on a retina MacBook to create and issue.

ChALkeR:
I rebased and performed minor style fixes, landing after CI.

ChALkeR:
Landed in 222211cb, thanks!

PR #358, opened 2016-11-10
engine/properties: Support properties starting with "on"

stephenmdangelo:
Fixes an exception for properties named "on"

No longer check for a signal for properties like "on_" (since
"_".toUpperCase() === "_")

If no signal is found, try to find a property with the name
(for properties like "onProp" that match the RegExp).


ChALkeR:
Landed in aa78402, thanks! I also did some minor code cleanup in 694ecdc.

PR #363, opened 2016-12-07
CustomElements: initial Custom Elements v1 support

ChALkeR:
This introduces support for registering QmlWeb based components as Custom Elements v1.

Refs:
 * https://www.w3.org/TR/custom-elements/,
 * https://developers.google.com/web/fundamentals/getting-started/primers/customelements.

One could now define QML Custom elements with the following call (given that the browser supports Custom Elements v1 or that a polyfill is loaded):

```js
QmlWeb.registerElement('qml-main', 'qml/main.qml');
```

And use them as:

```html
<qml-main width="300" height="300" color="red" firstName="World"></qml-main>
```

HTML attributes are constructed from the QML file, bound and dynamically updated, also updating HTML attributes has a direct effect on the QML state.

Example included, with a polyfill.

Caveat: atm, QmlWeb supports only one running engine, which limits the number of active elements. That has to be fixed later (see #364).

/cc @akreuzkamp @Plaristote @stephenmdangelo 

CI is red because of the decreased code coverage, I still need tests for this.

### DIFF
--- a/src/engine/CustomElements.js
+++ b/src/engine/CustomElements.js
@@ -1,0 +1,150 @@
+const anchorNames = [
+  "left", "right", "top", "bottom", "verticalCenter", "horizontalCenter"
+];
+
+const ignoreProps = [
+  "x", "y", "z", "scale", "rotation", "implicitWidth", "implicitHeight"
+];
+
+function getProperties(file) {
+  // TODO: implement a cleaner way
+
+  const div = document.createElement("div");
+  const engine = new QmlWeb.QMLEngine(div);
+  engine.loadFile(file);
+
+  const qml = engine.rootObject;
+  const properties = Object.keys(qml.$properties).filter(name => {
+    // Invalid names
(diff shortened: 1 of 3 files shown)


src/engine/CustomElements.js
```diff
@@ -0,0 +1,142 @@
...
+      // TODO: use something smarter than .offsetHeight?
+      div.style.height = `${this.offsetHeight}px`;
+
+      const engine = this.engine = new QmlWeb.QMLEngine(div);
+      engine.loadFile(file);
+      engine.start();
+      const qml = this.qml = engine.rootObject;
+
+      // Bind attributes
+      for (const attr of attributes) {
```
ChALkeR: Should probably avoid `for…of` for the time being, as transpliterated code for that is quite heavy and uses `Symbol.iterator`.

ChALkeR:
Going to go ahead and land this soon, we could fix all issues later if they arise.
This is a new feature so it won't break existing things anyway.

Issue #366, opened 2016-12-08
Dynamic elements sizing


ChALkeR:
We should support dynamic sizing of the elements, which is now supported only for the `body` element.

E.g. when a user initializes QmlWeb on a `div` the content should resize with that div if it doesn't have explicit size.

This is also probably needed for Web Components.

ChALkeR:
Fixed in b05ac999b811e96f54a96847ba55172d14db50f2.

Issue #365, opened 2016-12-07
Utilize Shadow DOM v1 to isolate QML elements


ChALkeR:
Shadow DOM v1 is out, and browser vendors seem to have agreed to implement that.

We should use Shadow DOM v1, when available, to isolate elements, so that:
 * QML child elements won't show up in `.querySelector()` and other DOM queries and won't pollute global tree (so they won't accidentially be affected by the user)
 * QML child elements won't be affected by the page CSS.

This will still allow easy debugging, as there is a switch in browsers that allows to inspect Shadow DOM contents.

When Shadow DOM is not available (as that depends on browser support), we should keep things in their current state.

Given that we will need to support operating both with and without Shadow DOM (to keep browsers support), we might also add a global flag that disables the isolation, if someone would want to keep the old behaviour and style QML child elements with css for some reason.

Refs:
 * https://www.w3.org/TR/shadow-dom/
 * https://developers.google.com/web/fundamentals/getting-started/primers/shadowdom

The patch against #363 is trivial:

```diff
diff --git a/src/engine/CustomElements.js b/src/engine/CustomElements.js
index aa9230b..e137829 100644
--- a/src/engine/CustomElements.js
+++ b/src/engine/CustomElements.js
@@ -50,7 +50,12 @@ function registerElement(name, file) {
       }
 
       const div = document.createElement("div");
-      this.appendChild(div);
+      if (true) { // TODO: check presence and add an opt-out
+        const shadow = this.attachShadow({mode: 'open'});
+        shadow.appendChild(div);
+      } else {
+        this.appendChild(div);
+      }
 
       // Initial height, overwritten in QML to computed height
       // TODO: use something smarter than .offsetHeight?
```

That introduces Shadow DOM only for child elements, I will think more about current `.loadFile(`/`.loadQML(` methods later.

/cc @akreuzkamp @Plaristote @stephenmdangelo

ChALkeR:
The patch against #363 is trivial:

```diff
diff --git a/src/engine/CustomElements.js b/src/engine/CustomElements.js
index aa9230b..e137829 100644
--- a/src/engine/CustomElements.js
+++ b/src/engine/CustomElements.js
@@ -50,7 +50,12 @@ function registerElement(name, file) {
       }
 
       const div = document.createElement("div");
-      this.appendChild(div);
+      if (true) { // TODO: check presense and add an opt-out
+        const shadow = this.attachShadow({mode: 'open'});
+        shadow.appendChild(div);
[...]

PR #367, opened 2016-12-09
[Breaking] Wrap all root elements

ChALkeR:
This is a breaking change, and it alters test expectations.

Before this change, all top-level QML elements were created in place of the DOM element passed to `QMLEngine(element)` constructor (let's call it «bounding element») and altered that element significantly, setting its width and height.

This PR only slightly changes the style of bounding element (ensuring `position` and `overflow: hidden`) and places the top-level QML DOM element **inside** the bounding element instead.

That allows to preserve the implicit sizing of the bounding `element` when needed, and QML element sizes are now be recalculated in runtime.

Bounding element size now always sets the top-level QML element `width` and `height` (instead of `implicitWidth` and `implicitHeight`), as it should.

Also the bounding element contents are not replaced now, but preserved.

Commit messages are not yet final and subject to change (I will explain things better and copy description from this PR). I will also squash the two commits here, they are split to make the review easier (tests changes are put into a separate commit for now).

This utilizes `window.getComputedStyle`, which [is available](http://caniuse.com/getComputedStyle) on everything newer than IE8, which we already don't support.

~Also, Shadow DOM v1 support is introduced, which allows us to hide and isolate the implementation so that it doesn't get affected by css styles. It is used only when supported and could be disabled either globally or per QMLEngine.~

_Upd: removed Shadow DOM support from this PR, because it broke implicit `Text` size calculation. I will file Shadow DOM support as a separate PR. Extracted to #368._

Fixes: ~#365,~ #366.
This will also allow to simplify #363.

/cc @akreuzkamp @Plaristote @stephenmdangelo

### DIFF
--- a/examples/embed.html
+++ b/examples/embed.html
@@ -4,7 +4,7 @@
     <title>QML Embed Example</title>
     <style>
       #embed {
-        width: 400px;
+        width: 50%;
         height: 300px;
         border: 10px dotted red;
       }
--- a/src/engine/QMLEngine.js
+++ b/src/engine/QMLEngine.js
@@ -14,7 +14,7 @@ class QMLEngine {
     this.fps = 60;
     // Math.floor, causes bugs to timing?
     this.$interval = Math.floor(1000 / this.fps);
-    this.rootElement = element;
+    this.dom = element || document.body;
 
     // Cached component trees (post-QmlWeb.convertToEngine)
     this.components = {};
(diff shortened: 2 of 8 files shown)


ChALkeR:
Ok, I'm pretty sure we need this, and this blocks other pull requests, so I'm going to go forward and land this.

If anyone has any objections — please just file an PR or issue afterwards.

Issue #371, opened 2016-12-23
Loader should inherit parent scope


ChALkeR:
See the failing testcase `Render/Simple/LoaderScope` added in b400794f923da518dd5002c40fc6580f1f598c91.

/cc @qmlweb/collaborators

ChALkeR:
Proposed fix in #373.

PR #373, opened 2016-12-26
Loader: fix children context

ChALkeR:
Fixes: https://github.com/qmlweb/qmlweb/issues/371

This works, I'm mostly sure that it was intended to be that way. Not 100%, though — I'm not very familiar with the contexts and scopes in QmlWeb.

It would be very nice if someone else could confirm this.

/cc @Plaristote @stephenmdangelo @akreuzkamp


stephenmdangelo:
LGTM, since I think all of the properties in the Loader will be in this.context as well.

ChALkeR:
@stephenmdangelo Thanks! =)

Issue #375, opened 2016-12-26
.pragma library support for js files


ChALkeR:
We definitely need that.

Work-around: include in one place, register to a property.
Another work-around: inside the library, bind the objects through `global` and unique (per-file) ids.

PR #377, opened 2016-12-26
Engine: support .pragma library for js files

ChALkeR:
Fixes: https://github.com/qmlweb/qmlweb/issues/375

Testcase is not ready yet.

/cc @stephenmdangelo for #255 and https://github.com/qmlweb/qmlweb-parser/pull/26 — perhaps you would be interested to try this out =).

### DIFF
--- a/src/engine/QMLEngine.js
+++ b/src/engine/QMLEngine.js
@@ -583,9 +583,24 @@ class QMLEngine {
       ${jsData.exports.map(sym => `$context.${sym} = ${sym};`).join("")}
     `);
 
-    this.js[file] = contextSetter;
+    if (jsData.pragma && jsData.pragma.indexOf("library") !== -1) {
+      // For .pragma library, execute the code once in a clean context and
+      // return a function that imports getters/setters to other contexts
+      const jsContext = {};
+      contextSetter(jsContext);
+      this.js[file] = context => {
+        jsData.exports.forEach(sym => QmlWeb.setupGetterSetter(context, sym,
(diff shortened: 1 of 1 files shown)


stephenmdangelo (changes requested):
I don't think this will match Qt QML behaviour for `Qt.include`, as that should create a copy of the included file, ignoring any pragma statements within it: http://doc.qt.io/qt-5/qtqml-javascript-imports.html#including-a-javascript-resource-from-another-javascript-resource

When importing a JavaScript file into a QML file, wouldn't it be faster/easier to keep a global reference to the context object that is created the first time it is imported (e.g. Whatever is assigned to the "as" variable, as opposed to the "contextSetter" function), and then create a variable with the qualified name inside of the Component context and assign the global object to it?

I think that would require changes outside of loadJS.

ChALkeR:
@stephenmdangelo Thanks.

I will check what could be done with `Qt.include`. I don't think that there are any performance issues here, at least not until the benchmarks have proven that, though.

I will try to do the other impl (like you proposed) and will compare the two patches on size and complexity (given that `Qt.include` support would be included into both).

stephenmdangelo:
I think for Qt.include, just maintain the old behaviour and ignore the pragma. That's why I think this feature has to go up a level or two to wherever  "importJavascriptInContext" is called. 

It's less the performance (unless a library has a million exports or something), and more just the simplicity. A global object per library that is just referenced by different names in the context of each QML file. 

In

PR #378, opened 2016-12-30
Fixed CustomeElement Support.

Gaubee:
due to `customElements.define` depend on ES6 code `class extends HtmlElement` which is not support compilation into ES5.
so use `document.registerElement` replace.

### DIFF
--- a/src/engine/CustomElements.js
+++ b/src/engine/CustomElements.js
@@ -56,95 +56,99 @@ function registerElement(name, file) {
     return map;
   }, {});
 
-  const QmlElement = class extends HTMLElement {
-    connectedCallback() {
-      // Default wrapper display is inline-block to support native width/height
-      const computedStyle = window.getComputedStyle(this);
-      if (computedStyle.display === "inline") {
-        this.style.display = "inline-block";
+  const qmlElementProto = Object.create(HTMLElement.prototype);
+
+  qmlElementProto.createdCallback = function() {
+    // Default wrapper display is inline-block to support native width/height
+    const computedStyle = window.getComputedStyle(this);
+    if (computedStyle.display === "inline") {
+      this.style.display = "inline-block";
+    }
+
+    const engine = this.engine = new QmlWeb.QMLEngine(this);
+    engine.loadFile(file);
+    engine.start();
+    const qml = this.qml = engine.rootObject;
+
+    // Bind attributes
+    attributes.forEach(attr => {
+      const pname = attr2prop[attr] || attr;
+      const val = this.getAttribute(attr);
+      if (typeof val === "string") {
+        qml[pname] = val;
       }
-
-      const engine = this.engine = new QmlWeb.QMLEngine(this);
-      engine.loadFile(file);
-      engine.start();
-      const qml = this.qml = engine.rootObject;
-
-      // Bind attributes
-      attributes.forEach(attr => {
-        const pname = attr2prop[attr] || attr;
-        const val = this.getAttribute(attr);
-        if (typeof val === "string") {
-          qml[pname] = val;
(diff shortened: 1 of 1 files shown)


ChALkeR:
No, things are a bit more complex here.

`document.registerElement` is not an alternative to `class extends HtmlElement`, it was present only in the previous version of the spec that won't get implemented by other browers (aka «[Custom Elementa v0](http://caniuse.com/custom-elements)»).

The spec was updated, `document.registerElement` was removed from the spec and replaced with `class extends HtmlElement`. That's what going to get implemented by all browsers, see «[Custom Elementa v1](http://caniuse.com/custom-elementsv1)».

There is no future for `document.registerElement` at this point.

Now, returning to QmlWeb — the current  implementation works when transplited into ES5 if and only if one includes a polyfill for Custom Elements, as could be seen in [/examples/customelement.html](https://github.com/qmlweb/qmlweb/blob/master/examples/customelement.html). That polyfill is required both to enable Custom Elements support in older browsers _and_ to patch the newer browsers so that they work fine with classes transplited to ES5.

If the user doesn't care about older browsers and wants to run only on native Custom Elements — they don't need transpliting to ES5 anyway (as browsers that support Custom Elements v1 support ES classes), so they could just disable the transpliter.

That could be done by removing the `es2015` preset from [.babelrc](https://github.com/qmlweb/qmlweb/blob/master/.babelrc) and rebuilding QmlWeb, for example.

It might be worth to consider providing un-transplited js files directly in the `/lib` directory, as older browsers usage share is decaying rapidly.

Gaubee:
For compatibility, How about wrap the ES6 source code into string to avoid babel compilation. And use `registerElement`  to transition.

ChALkeR:
I considered that at first, but that would require making a significant chunk of code less readable.

Long-term, it should be kept in the current state and the `class` transpliter should be disabled.

In fact, the only browser that stops us from dropping the `class` transplitter right now (IMO) is IE11. And I don't feel like we should lower the code readability only because of IE11.

Gaubee:
No only ie, my chrome 55 win10 x64 dropped too.
I think we could us plugin mode to include that .like as 'qmlweb.parser.js'

ChALkeR:
> No only ie, my chrome 55 win10 x64 dropped too.

No, I mean that of the noticeable browsers, only IE11 doesn't have `class` support. If we drop IE11, we can drop the transpliter for `class`es, and a compat shim won't be needed anymore.

> I think we could us plugin mode to include that .like as 'qmlweb.parser.js'

The reason for moving parser to a separate file was that it's huge and optional (not needed for pre-parsed files). The Custom Elements glueing code is small and while it is optional now, that might change later (if we choose to support declarative html element definitions, for example).

For me it looks like the [native-shim.js](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js) is the best approach here at the moment. Or does it cause any issues for your setup?

Gaubee:
Thanks.

I still have an question. About the es6 code without babel ,  could not work now.
Because the code use like 'class A' and 'superCall:A.call(this,meta)', but class could only use as 'new A'.

Why no use 'extends' and 'super(meta)' to replace?

ChALkeR:
@Gaubee Hm, thanks, I will take a look at that.

ChALkeR:
The problem here is that modules could have cross dependencies at this moment, this is why we are not using native es2015 class inheritance syntax — some types have to wait until the types that they depend on are loaded. I will take another look at that, perhaps there could be a better way.

ChALkeR:
@Gaubee Native ECMAScript 2015 class syntax is now live in master, along with `qmlweb.es2015.js` build target.

It will be packaged in 0.3.0.

PR #381, opened 2016-12-31
Timer: make parent not read-only for now

ChALkeR:
Some parts of QmlWeb set .parent property directly, that was causing a
conflict and an exception was being thrown.

Long-term, we should create a better fix, but this fixes the immediate
problem for now, and properties not being read-only when they should
causes much less damage than unexpected exceptions on valid qml code.

This still need a testcase.

### DIFF
--- a/src/modules/QtQml/Timer.js
+++ b/src/modules/QtQml/Timer.js
@@ -5,7 +5,7 @@ QmlWeb.registerQmlType({
   baseClass: "QtObject",
   properties: {
     interval: { type: "int", initialValue: 1000 },
-    parent: { type: "QtObject", readOnly: true },
+    parent: { type: "QtObject" }, // TODO ro
     repeat: "bool",
     running: "bool",
     triggeredOnStart: "bool"


stephenmdangelo:
I have this fix on a branch of mine that solves the issue for QObject.$delete. Are there other places you're seeing an exception being thrown for Timer?

https://github.com/stephenmdangelo/qmlweb/commit/d20d8204b7bb6b098396cfce114d4a13a919e691

ChALkeR:
@stephenmdangelo Thanks! I will test that commit soon (hopefully today) and will tell you if that works for my original testcase that was failing (it's not a minimal testcase yet).

ChALkeR:
@stephenmdangelo Full grep:
```js
src/modules/QtQuick/Item.js:        child.parent = this; // This will also add it to children.
src/modules/QtQuick/Repeater.js:    this.parent = meta.parent;
src/modules/QtQuick/Repeater.js:      newItem.parent = this.parent;
src/modules/QtQuick/Loader.js:    qmlComponent.parent = this;
src/modules/QtQuick/Loader.js:    this.item.parent = undefined;
src/modules/QtQuick.Controls/ScrollView.js:      newItem.parent = this;
src/qtbase/QObject.js:    this.parent = undefined;
```

PR #383, opened 2017-01-03
loadImports: Handle moduleVersion being null

stephenmdangelo:
This happens with pre-parsed files, as "undefined" is serialized to "null"
via JSON.

### DIFF
--- a/src/engine/modules.js
+++ b/src/engine/modules.js
@@ -179,7 +179,7 @@ function loadImports(self, imports) {
   }
   for (let i = 0; i < imports.length; ++i) {
     const [, moduleName, moduleVersion, moduleAlias] = imports[i];
-    if (typeof moduleVersion === "undefined") continue;
+    if (typeof moduleVersion !== "number") continue;
     const versionString = moduleVersion % 1 === 0 ?
                             moduleVersion.toFixed(1) :
                             moduleVersion.toString();


stephenmdangelo:
TODO: Add some pre-parsed qrc files to the tests.

src/engine/modules.js
```diff
@@ -179,7 +179,9 @@ function loadImports(self, imports) {
   }
   for (let i = 0; i < imports.length; ++i) {
     const [, moduleName, moduleVersion, moduleAlias] = imports[i];
-    if (typeof moduleVersion === "undefined") continue;
+    if (typeof moduleVersion === "undefined" || moduleVersion === null) {
```
ChALkeR: Perhaps just `!moduleVersion` (that would probably break module versions `0.0`, are those premitted in Qt?).

ChALkeR: Yes, it looks like `import Something 0` is permitted.

Perhaps `typeof moduleVersion !== 'number'`? That should be easier to read, and the code below relies on `moduleVersion` to be a `number`.

stephenmdangelo: Should be good now.

ChALkeR:
Landed in ddbd056, thanks!

stephenmdangelo:
I think you just merged this as a no-op a second time :)

ChALkeR:
Hm, it was open for some reason. Let me check =).

ChALkeR:
Um, yes, it's indeed no-op. Not sure why it was open on GitHub =).
Force-push time!

ChALkeR:
Fixed. Ok, this was landed in 2c893646693591a8d6e8737b27ac3cd9166429d6 =).

ChALkeR:
Fixed. Ok, this was landed in 2c893646693591a8d6e8737b27ac3cd9166429d6 =).

PR #384, opened 2017-01-06
Create QtGraphicalEffects:RectangularGlow.js

Gaubee:
use css `box-shadow` and `filter:blur` to simulate **Glow** effect.
They are very similar, but not exactly the same.

## REFERENCES:
[qml-qtgraphicaleffects-rectangularglow](http://doc.qt.io/qt-5/qml-qtgraphicaleffects-rectangularglow.html)
[css-box-shadow](https://developer.mozilla.org/zh-CN/docs/Web/CSS/box-shadow)

## Screenshots
| prop | qmlweb | qml |
|:----:|------|------|
|  | Chrome | Qt Creator Design Mode, with auxiliary dotted line |
|cornerRadius: 0; spread: 0; glowRadius: 20; color: "#ffffff";|image|image|
|cornerRadius: 50; spread: 0; glowRadius: 20; color: "#ffffff";|image|image|
|cornerRadius: 25; spread: 0; glowRadius: 10; color: "#ffffff";|image|image|
|cornerRadius: 25; spread: 0; glowRadius: 40; color: "#ffffff";|image|image|
|cornerRadius: 25; spread: 0; glowRadius: 20; color: "#ffffff";|image|image|
|cornerRadius: 25; spread: 1; glowRadius: 20; color: "#ffffff";|image|image|
|cornerRadius: 25; spread: 0.5; glowRadius: 20; color: "#55ff55";|image|image|

### DIFF
--- a/src/modules/QtGraphicalEffects/RectangularGlow.js
+++ b/src/modules/QtGraphicalEffects/RectangularGlow.js
@@ -60,20 +60,20 @@ QmlWeb.registerQmlType({
   $updateBoxShadow() {
 
     function calcBoxShadow(color, glowR, cornerR, spread) {
-      let totle = glowR + cornerR * (1 - spread);
-      let glow = (1 - spread) * totle;
-      let blur_radius = glow * 0.64;
-      let spread_radius = totle - blur_radius;
-      let glow2 = glowR / 5;
-      let blur_radius_2 = glow2 * 0.8;
-      let spread_radius_2 = glow2 - blur_radius_2;
+      const totle = glowR + cornerR * (1 - spread);
+      const glow = (1 - spread) * totle;
+      const blur_radius = glow * 0.64;
+      const spread_radius = totle - blur_radius;
+      const glow2 = glowR / 5;
+      const blur_radius_2 = glow2 * 0.8;
+      const spread_radius_2 = glow2 - blur_radius_2;
       return `${color} 0px 0px ${blur_radius}px ${spread_radius}px,` +
         `${color} 0px 0px ${blur_radius_2}px ${spread_radius_2}px`;
     }
 
     function calcGlowCss(color, glowR, cornerR, spread, width, height) {
-      let spread_cornerR = cornerR * (1 - spread);
-      let rest_cornerR = cornerR - spread_cornerR;
+      const spread_cornerR = cornerR * (1 - spread);
+      const rest_cornerR = cornerR - spread_cornerR;
       return {
         boxShadow: calcBoxShadow(color, glowR, cornerR, spread),
         width: `${width - spread_cornerR}px`,
--- a/src/modules/QtGraphicalEffects/RectangularGlow.js
+++ b/src/modules/QtGraphicalEffects/RectangularGlow.js
@@ -1,0 +1,92 @@
+QmlWeb.registerQmlType({
+  module: "QtQuick",
+  name: "RectangularGlow",
+  versions: /.*/,
+  baseClass: "Item",
+  properties: {
+    cached: {
+      type: "bool",
+      initialValue: false
+    },
+    color: {
+      type: "color",
+      initialValue: "white"
+    },
+    cornerRadius: {
+      type: "real",
+      initialValue: 0
+    },
+    // This property defines how many pixels outside the item area are reached by the glow.
+    glowRadius: {
+      type: "real",
+      initialValue: 0
+    },
+    spread: {
+      type: "real",
+      initialValue: 0
+    },
+  }
+}, class {
+  constructor(meta) {
+    QmlWeb.callSuper(this, meta);
+
+    const bg = this.impl = document.createElement("div");
+    bg.style.pointerEvents = "none";
+    bg.style.position = "absolute";
+    bg.style.left = bg.style.right = bg.style.top = bg.style.bottom = "0px";
+    bg.style.border = "none";
+    bg.style.backgroundColor = "black";
+    this.dom.appendChild(bg);
+
+    this.colorChanged.connect(this, this.$onColorChanged);
+    this.glowRadiusChanged.connect(this, this.$updateBoxShadow);
+    this.cornerRadiusChanged.connect(this, this.$updateBoxShadow);
+    this.widthChanged.connect(this, this.$updateBoxShadow);
+    this.heightChanged.connect(this, this.$updateBoxShadow);
+    this.spreadChanged.connect(this, this.$onSpreadChanged);
+  }
+  $onColorChanged(newVal) {
+    this.impl.style.backgroundColor = new QmlWeb.QColor(newVal);
+    this.$updateBoxShadow();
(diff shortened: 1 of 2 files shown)


Gaubee:
well, travis-ci checks failed due to code format.
How to avoid these problems in the local development?

stephenmdangelo:
@Gaubee You can run `npm test` before pushing to ensure everything passes.

Also, for some simple formatting, `eslint` can be run on a file to automatically fixup things like spacing:

```
eslint --fix src/file/that/failed.js
```

When running `eslint --fix`, I usually get some errors like this, but the fixes are still applied to the file:

```
  1:1  error  Definition for rule 'func-call-spacing' was not found            func-call-spacing
  1:1  error  Definition for rule 'no-global-assign' was not found             no-global-assign
  1:1  error  Definition for rule 'no-tabs' was not found                      no-tabs
  1:1  error  Definition for rule 'no-template-curly-in-string' was not found  no-template-curly-in-string
  1:1  error  Definition for rule 'no-unsafe-negation' was not found           no-unsafe-negation
```

ChALkeR:
@stephenmdangelo 
> When running `eslint --fix`, I usually get some errors like this, but the fixes are still applied to the file:

Just update your `eslint` version, those missing rules errors are most probably due to an outdated `eslint` version that you have installed locally. I belive that the eslint version specified in `package.json` should be fine (`^3.3.1`).

src/modules/QtGraphicalEffects/RectangularGlow.js
```diff
@@ -0,0 +1,95 @@
...
+        top: `${spread_cornerR / 2}px`,
+        left: `${spread_cornerR / 2}px`,
+        filter: `blur(${spread_cornerR / 2}px)`,
+        borderRadius: `${rest_cornerR / 2}px`,
+        transform: `scale(${(width - spread_cornerR / 4) / width},` +
+          `${(height - spread_cornerR / 4) / height})`,
+      };
+    }
+
+    Object.assign(this.impl.style, calcGlowCss(
```
ChALkeR: This [breaks IE11](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Browser_compatibility), which is otherwise should be working afaik.
Could you use direct assigment here instead? I.e.:
```js
const style = this.impl.style;
style.boxShadow = calcBoxShadow(color, glowR, cornerR, spread);
style.width = `${width - spread_cornerR}px`;
…
```

src/modules/QtGraphicalEffects/RectangularGlow.js
```diff
@@ -0,0 +1,95 @@
...
+      const blur_radius = glow * 0.64;
+      const spread_radius = totle - blur_radius;
+      const glow2 = glowR / 5;
+      const blur_radius_2 = glow2 * 0.8;
+      const spread_radius_2 = glow2 - blur_radius_2;
+      return `${color} 0px 0px ${blur_radius}px ${spread_radius}px,` +
+        `${color} 0px 0px ${blur_radius_2}px ${spread_radius_2}px`;
+    }
+
+    function calcGlowCss(color, glowR, cornerR, spread, width, height) {
```
ChALkeR: A closure/return isn't needed here, you could do this instead:
```js
const { color, glowR, cornerR, spread, width, height } = this;
const spread_cornerR = cornerR * (1 - spread);
const rest_cornerR = cornerR - spread_cornerR;
```

This will also help with removing the `Object.assign` below.

src/modules/QtGraphicalEffects/RectangularGlow.js
```diff
@@ -0,0 +1,95 @@
+QmlWeb.registerQmlType({
+  module: "QtQuick",
+  name: "RectangularGlow",
+  versions: /.*/,
+  baseClass: "Item",
+  properties: {
+    cached: {
+      type: "bool",
+      initialValue: false
```
ChALkeR: Techically, `initialValue` is not required here and below (except for the `color`) — if `initialValue` is not specified, a type-specific default is used instead (like in QML). For `bool`, that is `false`, for numeric — `0`, for `color` — black.

It's ok to include those, though, so you can ignore this comment.

src/modules/QtGraphicalEffects/RectangularGlow.js
ChALkeR: Ah, to clarify the IE11 note — I don't expect `filter` to be displaying the correct thing on IE11, I just expect this to now throw =). Graceful degradation for various browsers is fine, throwing probably isn't.

Gaubee:
Any changes request?

use `Blur` is a relatively simple method. The difference between `Glow` and `Shadow` does not have a specific formula for transformation. `Blur` seems to be the best way to save code. But bring some compatibility issues.

ChALkeR:
@Gaubee I landed three commits with minor fixes/cleanups on your branch, could you take a look and check if everything is ok?

src/modules/QtGraphicalEffects/RectangularGlow.js
```diff
@@ -0,0 +1,92 @@
...
+
+    this.colorChanged.connect(this, this.$onColorChanged);
+    this.glowRadiusChanged.connect(this, this.$updateBoxShadow);
+    this.cornerRadiusChanged.connect(this, this.$updateBoxShadow);
+    this.widthChanged.connect(this, this.$updateBoxShadow);
+    this.heightChanged.connect(this, this.$updateBoxShadow);
+    this.spreadChanged.connect(this, this.$onSpreadChanged);
+  }
+  $onColorChanged(newVal) {
+    this.impl.style.backgroundColor = new QmlWeb.QColor(newVal);
```
ChALkeR: `new QmlWeb.QColor` is not required, `newVal` is already an instance of `QColor` here.

src/modules/QtGraphicalEffects/RectangularGlow.js
```diff
@@ -41,7 +31,7 @@ QmlWeb.registerQmlType({
     this.spreadChanged.connect(this, this.$onSpreadChanged);
   }
   $onColorChanged(newVal) {
-    this.impl.style.backgroundColor = new QmlWeb.QColor(newVal);
+    this.impl.style.backgroundColor = newVal;
```
Gaubee: newVal.$css? If #386 is merged at the same time.

ChALkeR: Yes, this should be converted to `newVal.$css` in some of the PRs, depending on which one of those would be merged first. I will handle that while landing.

ChALkeR:
It looks like #386 still needs some fixes, but this one looks good. I will handle the `.$css` change in the #386.

@Gaubee Is there anything else to do here, or is this ready to land?

Gaubee:
ready to land.

ChALkeR:
@Gaubee Landed in b115743, thanks!

PR #385, opened 2017-01-07
Create ColorAnimation.js

Gaubee:
This file include color parser to support color format like `#33ffffff`;

## Example
```qml
    Rectangle {
        color: "black"
        ColorAnimation on color { 
            to: "#33ffffff"
            duration: 10000
            running: true
        }
    }
```

shoud I move color parser to `QColor.js`?


Gaubee:
How to pass **codecov checks**?

ChALkeR:
@Gaubee 
> How to pass **codecov checks**?

By including tests for the new methods. At least, the new classes need entries in the [initialization tests](https://github.com/qmlweb/qmlweb/blob/master/tests/Initialize/runner.js#L9) — but that checks only that `ColorAnimation {}` does not throw on instantiation.


Regarding this PR, some of the functionality present here has to go into other files, e.g. the color parsing should probably go into [QColor.js](https://github.com/qmlweb/qmlweb/blob/master/src/qtbase/QColor.js).

Example (which now fails):
```qml
import QtQuick 2.0

Item {
    property color foo
    property color bar: 'green'
    Text { text: foo  + ' ' + bar }
}
```

should render `#000000 #008000`, which is currently not implemented.

Gaubee:
Well, I have a proposal to use the canvas to get the RGB value of a color already named. But I do not know the relevance of the compatibility of the named colors in modern browsers is there a uniform standard?

ChALkeR:
I am not sure why would we want `canvas` here, we just need to support the colors in the Qt format, as documented here: http://doc.qt.io/qt-5/qml-color.html.

Qt supports SVG color names: https://www.w3.org/TR/SVG/types.html#ColorKeywords

This is the CSS Colors specification — that is what the browsers use: https://www.w3.org/TR/css3-color/

Gaubee:
@ChALkeR use canvas can reduce the code, but not necessarily. This is not a good idea, given the problem of increased initialization time.

----

qml color `toString` should return **ARGB**(like `#FF000000`), but for css should be **RGBA**(like `rbga(0,0,0,1)`).

It will affect a lot of existing code.
For example:
```js
this.impl.style.backgroundColor = new QmlWeb.QColor(newVal);
//====>
const color = new QmlWeb.QColor(newVal);
this.impl.style.backgroundColor = color.$value_css;
```
What is the best practice you think.

ChALkeR:
@Gaubee I added an alternate impl that I started working on when I mentioned that QColor.js should be fixed. It's in the #386 PR, could you take a look if that does the thing?

src/qtbase/QColor.js
```diff
@@ -1,17 +1,208 @@
...
+    .match(/[A-Z][0-9|a-z]+/g);
```
ChALkeR: There is no reason to do something like this, what's wrong with a regular object?

Gaubee: Just to save the number of lines of code, with low-cost initialization time

Of course, regular objects well.

Of course, regular objects well.

Gaubee:
#386

ChALkeR:
@Gaubee No, let's reopen this, #386 just makes QColor implementation more compliant, it doesn't add ColorAnimation. Perhaps you could rebase this after #386 lands?

Gaubee:
@ChALkeR yeah. I will create new pull request after #386 lands.

ChALkeR:
#386 landed.

PR #386, opened 2017-01-09
QColor: complete implementation

ChALkeR:
This implements toString that returns results similar to what Qt QML does.
Also, Qt.hsla() and Qt.hsva() methods were fixed to return a QColor.

Css color specification is now received as .$css property.

New testcases are included.

/cc @Gaubee, @stephenmdangelo

### DIFF
--- a/src/qtbase/QColor.js
+++ b/src/qtbase/QColor.js
@@ -1,32 +1,205 @@
-// TODO complete implementation (with attributes `r`,`g` and `b`).
-
 class QColor {
-  constructor(val) {
-    this.$value = "black";
-    if (val instanceof QColor) {
+  constructor(...args) {
+    this.$changed = new QmlWeb.Signal();
+    this.$r = this.$g = this.$b = 0;
+    this.$a = 1;
+    const val = args[0];
+    if (args.length >= 3) {
+      this.$r = args[0];
+      this.$g = args[1];
+      this.$b = args[2];
+      if (args.length >= 4) {
+        this.$a = args[3];
+      }
+    } else if (val instanceof QColor) {
       // Copy constructor
-      this.$value = val.$value;
+      this.$a = val.a;
+      this.$r = val.r;
+      this.$g = val.g;
+      this.$b = val.b;
     } else if (typeof val === "string") {
-      this.$value = val.toLowerCase();
+      const lval = val.toLowerCase();
+      if (QColor.colormap[lval]) {
+        const rgb = QColor.colormap[lval];
+        this.$r = rgb[0] / 255;
+        this.$g = rgb[1] / 255;
+        this.$b = rgb[2] / 255;
+      } else if (lval === "transparent") {
+        this.$a = 0;
+      } else if (lval[0] === "#") {
+        const hex = lval.substr(1);
+        if (hex.length === 3) {
+          this.$r = parseInt(hex[0], 16) / 15;
+          this.$g = parseInt(hex[1], 16) / 15;
+          this.$b = parseInt(hex[2], 16) / 15;
+        } else {
+          const rgb = hex.match(/.{2}/g).map(x => parseInt(x, 16));
+          if (rgb.length === 4) {
+            this.$a = rgb.shift() / 255;
+          }
+          this.$r = rgb[0] / 255;
+          this.$g = rgb[1] / 255;
+          this.$b = rgb[2] / 255;
+        }
+      } else {
+        throw new Error(`Can not convert ${val} to color`);
+      }
     } else if (typeof val === "number") {
       // we assume it is int value and must be converted to css hex with padding
-      const rgb = (Math.round(val) + 0x1000000).toString(16).substr(-6);
-      this.$value = `#${rgb}`;
+      const hex = (Math.round(val) + 0x1000000).toString(16).substr(-6);
+      const rgb = hex.match(/.{2}/g).map(x => parseInt(x, 16));
+      this.$r = rgb[0] / 255;
+      this.$g = rgb[1] / 255;
+      this.$b = rgb[2] / 255;
     }
   }
   toString() {
-    return this.$value;
+    if (this.$string) return this.$string;
+    const argb = [this.$a, this.$r, this.$g, this.$b].map(x =>
+      (Math.round(x * 255) + 0x100).toString(16).substr(-2)
+    );
+    if (argb[0] === "ff") {
+      argb.shift(); // We don't need alpha if it's ff
+    }
+    this.$string = `#${argb.join("")}`;
+    return `#${argb.join("")}`;
+  }
+  get $css() {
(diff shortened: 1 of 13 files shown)


ChALkeR:
As for the default `a = 1` in `Qt.rgba`, `Qt.hsva`, and `Qt.hsla` — that's how Qt QML behaves, judging by the quick tests all those methods have a three-argument version with the default alpha equal to 1.

tests/QtBase/QColor.js
```diff
@@ -15,6 +15,43 @@ describe("QtBase.QColor", function() {
     });
   });
 
+  it("toString", function() {
+    var colors = [
+      ["transparent", "#00000000"],
+      ["#abcDEF", "#abcdef"],
+      ["red", "#ff0000"],
+      ["#01234567", "#01234567"]
```
ChALkeR: Perhaps I should add `["#18d", "#1188dd"]` here to test short hex colors.

src/qtbase/QColor.js
```diff
@@ -1,32 +1,78 @@
-// TODO complete implementation (with attributes `r`,`g` and `b`).
-
 class QColor {
   constructor(val) {
+    this.r = this.g = this.b = 0;
```
Gaubee: Why no save `r g b` as 0~255 replace 0~1 in constructor. 
For readability and could be faster than `Math.round(v * 255);` in function `$css` and `toString`.

src/qtbase/QColor.js
```diff
@@ -1,32 +1,78 @@
...
-      const rgb = (Math.round(val) + 0x1000000).toString(16).substr(-6);
-      this.$value = `#${rgb}`;
+      const hex = (Math.round(val) + 0x1000000).toString(16).substr(-6);
+      const rgb = hex.match(/.{2}/g).map(x => parseInt(x, 16));
+      this.r = rgb[0] / 255;
+      this.g = rgb[1] / 255;
+      this.b = rgb[2] / 255;
     }
   }
   toString() {
```
Gaubee: you could cache result. 
Especially the $ css function. can bring significant performance improvements in the animation.

src/qtbase/QColor.js
ChALkeR: Because QML `color` type actually does have `r`, `g`, `b`, and `a` properties, and those are in 0-1 range:

```qml
import QtQuick 2.0

Item {
    property color bar: 'green'
    Component.onCompleted: {
      console.log(bar.r, bar.g, bar.b, bar.a, bar)
    }
}
```

ChALkeR: Those are also settable:
```qml
import QtQuick 2.0

Item {
    property color bar: 'green'
    Component.onCompleted: {
      console.log(bar.r, bar.g, bar.b, bar.a, bar);
      bar.r = 1;
      bar.a = 0.5;
      console.log(bar.r, bar.g, bar.b, bar.a, bar);
    }
}
```

ChALkeR:
@Gaubee as for the caching stringified values — I considered that, that would require making `r`/`g`/`b`/`a` properties getters/setters that reset the cached value.

I was under an impression that this is still going to be fast, and that adding caching will just slow things down due to setters.

It would help if you could provide any specific benchmarks that are related to the actual usage patterns (e.g. animations).

ChALkeR:
Note: this doesn't implement `*Changed` signals yet when the color is modified through settings its `r`/`g`/`b`/`a` properties.

Testcase:
```qml
import QtQuick 2.0

Rectangle {
  property color bar: 'green'
  Component.onCompleted: {
    bar.r = 1;
    bar.a = 0.5;
  }
  onBarChanged: console.log(bar)
}
```

That isn't implemented for any of the basic types yet, and that would require making those properties getters/setters anyway, so perhaps caching stringified values would come at almost zero cost here.

src/qtbase/QColor.js
ChALkeR: Also notice how the internal `color` representation has more r/g/b/a steps than 256 — those look like floats, which could be important for hsla/hsva representation and `Qt.ligther`/`Qt.darker` (but those are not implemented yet).

Gaubee: Since animations require frequent invocation of the `$css` property, I recommend setting rgb to the calculation property for performance reasons. With the value cache, you can significantly improve performance.
```js
    class QColor{
        get r() {
            return this._r
        }
        set r(v) {
            this.toString = this._toStringWithCalc;
            this.get_css = this.$cssWithCalc;
            this._r = v;
        }
        toStringWithCalc() {
            this.toSting = this._toStringFromCache
        }
        toStringFromCache() {}
        $cssWithCalc() {
            this.get_css = this.$cssFromCache
        }
        $cssFromCache() {}
        get $css() {
            return this.get_css();
        }
    }
```

ChALkeR:
This should also implement `hsvHue`, `hsvSaturation`, `hsvValue`, `hslHue`, `hslSaturation`, `hslLightness` settings.

ChALkeR:
This should also implement `hsvHue`, `hsvSaturation`, `hsvValue`, `hslHue`, `hslSaturation`, `hslLightness` getters.

ChALkeR:
I added setters, cache, and `$changed` signal, but that signal is not yet connected to QML properties.
Not all methods are implemented yet.

ChALkeR:
Ok, this looks to be finished now. I will only add some more testcases before landing.

@Gaubee, does this suit for `ColorAnimation` needs?

src/qtbase/QColor.js
```diff
@@ -1,32 +1,204 @@
...
+    const argb = [this.$a, this.$r, this.$g, this.$b].map(x =>
+      (Math.round(x * 255) + 0x100).toString(16).substr(-2)
+    );
+    if (argb[0] === "ff") {
+      argb.shift(); // We don't need alpha if it's ff
+    }
+    this.$string = `#${argb.join("")}`;
+    return `#${argb.join("")}`;
+  }
+  get $css() {
```
ChALkeR: I should probably add a note here that `$css` will be removed long-term.

Modern browsers support color specification in `#rgba` and `#rrggbbaa` format, like Qt uses and like `toString()` above returns.

Refs: https://www.chromestatus.com/feature/5685348285808640
Refs: https://drafts.csswg.org/css-color/#hex-notation

ChALkeR: Argh, the problem is that the browsers use `#rgba` and Qt uses `#argb`, so that won't work.

Gaubee:
Wait for lands .
I will try optimize performance for coloranimate.

ChALkeR:
Setters cause colors to be not comparable in Qt: https://bugreports.qt.io/browse/QTBUG-58147
Will wait for that to get fixed upstream before adding setter tests.

ChALkeR:
Landed in 14f6e5968896af25ae0d59e659d9ef94552e6be5.

PR #390, opened 2017-01-13
QColor: remove comparable colors hack

ChALkeR:
When colors were modified using setter properties, that hack caused all
colors that were equal to the current one also being modified.

Breaking ==/=== that have an alternative of using Qt.colorEqual is
better than breaking color property setters.

/cc @Gaubee

### DIFF
--- a/src/engine/QMLProperty.js
+++ b/src/engine/QMLProperty.js
@@ -135,10 +135,6 @@ class QMLProperty {
       );
     }
 
-    if (this.val && this.val.$get) {
-      return this.val.$get();
-    }
-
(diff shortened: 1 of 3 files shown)


ChALkeR:
This is going to land soon if noone objects.

Note: comparing colors in Qt QML also has issues under some circumstances: https://bugreports.qt.io/browse/QTBUG-58147

Also, `Qt.colorEqual` will still work here.

Gaubee:
Yeah, I think ColorAnimation can use a single instance of. QColor after this lands.

PR #391, opened 2017-01-14
Add ColorAnimation and Fix SequentialAnimation

Gaubee:
Fix SequentialAnimation `running, paused` property.
Signed-off-by: Gaubee <gaubeebangeel@gmail.com>


src/modules/QtQuick/ColorAnimation.js
```diff
@@ -0,0 +1,132 @@
...
+      const action = this.$actions[i];
+      const property = action.target.$properties[action.property];
+      const progress = this.easing.$valueForProgress(this.$at);
+
+      const {
+        from,
+        to
+      } = action;
+
+      const cur_color = new QColor( // eslint-disable-line no-undef
```
ChALkeR: `QmlWeb.QColor` and you don't need the comment =).

src/modules/QtQuick/ColorAnimation.js
```diff
@@ -0,0 +1,132 @@
...
+      }
+    }
+    this.$startLoop();
+  }
+  $startLoop() {
+    for (const i in this.$actions) {
+      const action = this.$actions[i];
+      const from_color = action.from_color ||
+        action.target[action.property] ||
+        EMPTY_ANIMATION_COLOR;
```
ChALkeR: Does it make sense to cache this in a separate variable? This isn't the place where the slowdown would be here.
1. This usecase (when neither `from` nor target property value) is specified is very unlikely.
2. This only happens on the beginning of each loop and saves at most of one `new` call.

I would just write `new QmlWeb.QColor()` here, I think that it wouldn't do any measurable difference, and that would keep the code a bit cleaner.

src/modules/QtQuick/ColorAnimation.js
```diff
@@ -0,0 +1,132 @@
...
+
+      const cur_color = new QmlWeb.QColor(
+        from.r + (to.r - from.r) * progress,
+        from.g + (to.g - from.g) * progress,
+        from.b + (to.b - from.b) * progress,
+        from.a + (to.a - from.a) * progress
+      );
+
+      // Notification changes
+      property.set(cur_color, QmlWeb.QMLProperty.ReasonAnimation);
```
ChALkeR: _Ignore this comment, just future optimization thoughts_.

This one, on the contrary, is called on each tick.
I wonder if this could be optimized…
Perhaps that could be done via creating a special QColor instance for each animation only at the start, and modifiying that with a special rgba-setter (that I could add, if needed).

Ignore this comment for now, I will think about optimizing that later myself (or just ignore that if future benchmarks will show that everything is ok here).

Note: QColor, unlike other basic types, is pretty optimized already — it doesn't create getters/setters for its properties in the runtime, those are all defined in the class itself.

ChALkeR: _Ignore this comment, just future optimization thoughts_.

This one, on the contrary, is called on each tick.
I wonder if this could be optimized…
Perhaps that could be done via creating a special QColor instance for each animation only at the start, and modifiying that with a special rgba-setter (that I could add, if needed).

Ignore this comment for now, I am not sure if that won't break anything, and will think about optimizing that later myself (or just ignore that if future benchmarks will show that everything is ok here).

Note: QColor, unlike other basic types, is pretty optimized already — it doesn't create getters/setters for its properties in the runtime, those are all defined in the class itself.

Gaubee: I had try to use a special QColor instance for each animation only at the $startLoop called.
But something wrong happened, due to [QMLProperty.js](https://github.com/qmlweb/qmlweb/blob/master/src/engine/QMLProperty.js#L191-L207) will reset  value.
I do not know where the specific error, so take the most cost-effective way to achieve.
This does not cause performance problems for Modern browser. For performance optimization, we have to focus on reducing the operation of the DOM.

-----

This is an TestCode:
```qml
import QtQuick 2.0

Item {
    width: 300
    height: 300
    Rectangle {
        id: rect
        anchors.fill: parent
        color: "red"

        SequentialAnimation {
            id:ani
            running:false
            ColorAnimation { target: rect; property: "color"; from: 'red'; to: 'green'; duration: 2000 }
            ColorAnimation { target: rect; property: "color"; to: 'blue'; duration: 1000 }
       }
        Text{
            text:'<b>'+parent.color+'</b>'
            anchors.centerIn: parent
        }
        MouseArea{
            onClicked:console.log(ani.running = !ani.running)
            anchors.fill: parent;
        }
    }
}
```
and the Timeline info:
image

you see, the $updateImplicit spent a lot of time. That is what we should focus.

Plaristote:
Alright, I merged these changes using cherry picks and manual labor.
c519da6c0c7e235b4a6da6bc1090e0178bf412a9
5faff85d3eb2d223a5cd836b87172f2d760ae6b9
0bec11f8c96e96b14504d554648b39ac2312010b
fe5b1eeafe077eba487107978f2aa5efd4e1b530

The tests are passing, but some issues remain.

- The start and complete colors are a match, so that's good... but the animation doesn't really seem to run.
I've been able to fix the issue, but I broke the tests doing so, so I'll be trying to fix that up before pushing it.

- ColorAnimation doesn't work on bound properties (when `from` or `to` are null and `oldVal`/`newVal` are used instead in the animations `$actions`). I've made a fix for that, but it also breaks the tests for now.

PR #393, opened 2017-01-15
Fix Scale default value.

Gaubee:


ChALkeR:
Landed in c782474, thanks!

PR #404, opened 2017-04-17
April fixes

pavelvasev:
Some fixes

### DIFF
--- a/src/modules/QtQuick.Controls/ComboBox.js
+++ b/src/modules/QtQuick.Controls/ComboBox.js
@@ -22,8 +22,15 @@ QmlWeb.registerQmlType({
     this.dom.style.pointerEvents = "auto";
     this.name = "QMLComboBox";
 
+    // TODO change innerHTML to DOM
+    this.dom.innerHTML = "<select></select>";
+    this.impl = this.dom.firstChild;
+
     this.Component.completed.connect(this, this.Component$onCompleted);
     this.modelChanged.connect(this, this.$onModelChanged);
+    this.currentIndexChanged.connect(this, this.$onCurrentIndexChanged);
+    this.heightChanged.connect(this, this.$onHeightChanged);
+    this.widthChanged.connect(this, this.$onWidthChanged);
 
     this.dom.onclick = () => {
       const index = this.dom.firstChild.selectedIndex;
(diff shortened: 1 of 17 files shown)


ChALkeR:
Uh-oh, this is a big one, thanks!

Will it be ok if I force-push to this branch to reorder and regroup things?
I made a copy of this branch to https://github.com/qmlweb/qmlweb/tree/april-fixes-orig to be sure that I don't miss anything.

src/modules/QtQuick/Image.js
```diff
@@ -15,6 +15,7 @@ QmlWeb.registerQmlType({
     asynchronous: { type: "bool", initialValue: true },
     cache: { type: "bool", initialValue: true },
     smooth: { type: "bool", initialValue: true },
+    noFlicker: { type: "bool", initialValue: false },
```
ChALkeR: This is not a standard property, so we shouldn't add it. We should stick to one of the behaviours by default to more accurately match the Qt behaviour.

The additional functionality could go somewhere else, but not in QtQuick module.
E.g. it could be something like `QmlWeb/Image` or `QmlWeb.Addons/Image`, though I doubt if we really need that — probably the best way would be to just do what Qt does by default.

src/modules/QtQuick.Controls/ComboBox.js
```diff
@@ -43,29 +49,72 @@ QmlWeb.registerQmlType({
...
-      //        html += "<option>" + elt[j] + "</option>";
-      //}
-      //else
-      entries.push(`<option>${elt}</option>`);
-    }
-    // TODO: remove innerHTML, port to DOM
-    this.dom.innerHTML = `<select>${entries.join("")}</select>`;
+
+    // TODO change innerHTML to DOM
+    this.dom.innerHTML = "<select></select>";
```
ChALkeR: This is nice. Now, constructing DOM via setting `innerHTML` is not needed here — I will remove that after merging.

src/modules/QtQuick.Controls/ComboBox.js
```diff
@@ -6,6 +6,8 @@ QmlWeb.registerQmlType({
   properties: {
     count: "int",
     currentIndex: "int",
+    size: { type: "int", initialValue: 1 },
+    // size is non-standard property to qtquick, but useful in dom
```
ChALkeR: Probably should go somewhere else, but not into the `QtQuick.Controls` module.
E.g. in `QmlWeb/ComboBox` or `QmlWeb.Addons/ComboBox`.

I would prefer if we kept compatibility both ways as much as possible for the modules that are defined in Qt.

.remarkrc
```diff
@@ -23,7 +23,7 @@
       "list-item-indent": "space",
       "list-item-spacing": false,
       "maximum-heading-length": 60,
-      "maximum-line-length": 80,
+      "maximum-line-length": 120,
```
ChALkeR: Is this needed in this PR?
I can't see any other change here that requires this one, there are no changes in the Markdown files in this PR.
So — why this change?

ChALkeR:
@pavelvasev re: line length in JS code — that's intended for the actual code.
I understand that on most monitors even 150 chars per line does not add a scrollbar, but the primary intention is to make code more readable.

Introducing a 80 chars per line limit for code actually forced making code more readable and splitting the logic by introducing extra variables and naming them accordingly.

E.g. these changes was forced by the 80 chars/line limit:
```diff
-          parts.push(`${this.$myEncodeURIComponent(key)}=${this.$myEncodeURIComponent(value)}`);
+          const ekey = this.$myEncodeURIComponent(key);
+          const evalue = this.$myEncodeURIComponent(value);
+          parts.push(`${ekey}=${evalue}`);
```
```diff
-      this.$parent.$tidyupList.splice(this.$parent.$tidyupList.indexOf(this), 1);
+      const index = this.$parent.$tidyupList.indexOf(this);
+      this.$parent.$tidyupList.splice(index, 1);
```

src/modules/QtQuick/Item.js
```diff
@@ -314,7 +314,7 @@ QmlWeb.registerQmlType({
   $onHeightChanged_(newVal) {
     this.css.height = newVal ? `${newVal}px` : "auto";
   }
-  $onFocusChanged(newVal) {
+  $onFocusChanged_(newVal) {
```
ChALkeR: Uh-oh. Thanks! =)

pavelvasev:
Hi @ChALkeR ! It sounds ok. Thank you very much for your management of the code.

src/modules/QtQuick.Controls/ComboBox.js
pavelvasev: ok dismiss it

.remarkrc
pavelvasev: ok dismiss it

src/modules/QtQuick/Image.js
pavelvasev: ok dismiss it

pavelvasev:
Hi @ChALkeR! I commited changes you note. Please review and merge..

ChALkeR:
@pavelvasev Sorry for the long delay on this, I got distracted by the [npm credentials](https://github.com/ChALkeR/notes/blob/master/Gathering-weak-npm-credentials.md) issue.

I will sort this out shortly.

ChALkeR:
Rebased against current `master`, working on this now.

Btw, re: `noFlicker` option — are there reaons why that is unwanted as the default behaviour?
How does Qt QML load those images?

ChALkeR:
About 50% done with sorting this out =).

ChALkeR:
Ok, I sorted this out.

@pavelvasev, some notes:

### 1. noFlicker option

Are there reasons why that is unwanted as the default behaviour?
How does Qt QML load those images?

### 2. Merged changes

I squashed them in several commits.

I rebased your original `april-fixes` branch against `master` to 570389b (aka `april-fixes-orig`).

Merged changes: d439145...4ed4386 (aka new `april-fixes`)

Ignored changes: 4ed4386...20425dc (aka `april-fixes-2` — mostly whitespace and noop). E.g. setting an `int` property default value to `0` does nothing — it's already 0 by default if not specified otherwise.

`april-fixes` + `april-fixes-2` are exactly equal to `april-fixes-orig` (which is your original branch, but rebased against current master).

ChALkeR:
Merged, thanks!

pavelvasev:
noFlicker option

Are there reasons why that is unwanted as the default behaviour?


Just the case. No problem.

23 июля 2017 г. 16:47 пользователь "Сковорода Никита Андреевич" <
notifications@github.com> написал:

pavelvasev:
Thank you, @Chalker!

23 июля 2017 г. 17:39 пользователь написал:

PR #418, opened 2017-09-14
1 div per Rectangle, Behaviors working smoothly, fix crash IE when re…

andresilvasantos:
1 div per Rectangle, Behaviors working smoothly, roll back registering class directly for Animation and Behavior - crashing on IE, added ColorAnimation (from Gaubee), added Grid column and row spacings, text font update - class QFont not emitting signal change, dont' print warning when ignoreUnkownSignals for Connections, TextInput without rect rendering, MouseArea press and release signals, added Item horizontal and vertical center offsets.

### DIFF
--- a/src/modules/QtQuick/MouseArea.js
+++ b/src/modules/QtQuick/MouseArea.js
@@ -9,13 +9,15 @@ QmlWeb.registerQmlType({
     hoverEnabled: "bool",
     mouseX: "real",
     mouseY: "real",
-    pressed: "bool",
+    isPressed: "bool",
     containsMouse: "bool",
     pressedButtons: { type: "variant", initialValue: 0 },
     cursorShape: "enum" // Qt.ArrowCursor
   },
   signals: {
     clicked: [{ type: "variant", name: "mouse" }],
+    pressed: [{ type: "variant", name: "mouse" }],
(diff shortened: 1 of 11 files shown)


src/modules/QtQuick/MouseArea.js
```diff
@@ -9,13 +9,15 @@ QmlWeb.registerQmlType({
     hoverEnabled: "bool",
     mouseX: "real",
     mouseY: "real",
-    pressed: "bool",
+    isPressed: "bool",
```
ChALkeR: Why this change? Property is named `pressed` in QML

src/modules/QtQuick/MouseArea.js
```diff
@@ -9,13 +9,15 @@ QmlWeb.registerQmlType({
...
     containsMouse: "bool",
     pressedButtons: { type: "variant", initialValue: 0 },
     cursorShape: "enum" // Qt.ArrowCursor
   },
   signals: {
     clicked: [{ type: "variant", name: "mouse" }],
+    pressed: [{ type: "variant", name: "mouse" }],
```
ChALkeR: Ah, to have an `onPressed` signal…

Hm, looks like there is a confict currently for elements that have properties and signals under the same name.
Will look into it.

Issue #436, opened 2018-05-12
singleton definition is not supported by readQmlDir


bessorion:
In accordance to http://doc.qt.io/qt-5/qtqml-modules-qmldir.html optional [singleton] keyword can be used for declaring singleton type, but current implementation of readQmlDir does not support more than 3 words in line. https://github.com/qmlweb/qmlweb/blob/9504d996a37cb8f50670b100b5cb51a2613edd7d/src/engine/import.js#L132

PR #437, opened 2018-05-13
Added failingTest for qmldir singleton difinition support

bessorion:
In accordance to http://doc.qt.io/qt-5/qtqml-modules-qmldir.html import statement can be declared as singleton. 
For now, this syntax is ignored and shows "unmatched" log message:
https://github.com/qmlweb/qmlweb/blob/9504d996a37cb8f50670b100b5cb51a2613edd7d/src/engine/import.js#L143
Added test covers #436


ChALkeR:
Thanks! I would prefer to migrate most tests to QtTest-based, but this format is fine for now.

PR #439, opened 2018-05-18
Add Yauhen Kotau to AUTHORS

ChALkeR:
Ref: https://github.com/qmlweb/qmlweb/pull/437

/cc @bessorion — could you confirm that the data is correct?


bessorion:
Yes, data is correct

PR #440, opened 2018-05-22
Properties and signals naming conflicts

bessorion:
Added failingTests to show naming conflicts between properties and signals. 
QML allows such behavior. Without this feature we cannot fully implement MouseArea QML type (and maybe some more items) which contains both property and signal with "pressed" names.
http://doc.qt.io/qt-5/qml-qtquick-mousearea.html


tests/QMLEngine/properties.js
```diff
@@ -129,4 +129,20 @@ describe("QMLEngine.properties", function() {
       QmlWeb.engine.$basePath + "remoteBinding.png");
     expect(qml.remoteSet).toBe(QmlWeb.engine.$basePath + "remoteSet.png");
   });
+
+  it("Property and signal can have the same names", function() {
+    var qml = load("AndSignalsWithTheSameNames", this.div);
+    //expect(qml.pressed).toBe(false);
```
ChALkeR: Why is this comment needed?

tests/QMLEngine/qml/PropertiesAndSignalsWithTheSameNames2.qml
```diff
@@ -0,0 +1,20 @@
+import QtQuick 2.5
+Item {
+  property bool signalCaught: false
+  property Item area: mouse_area
+  signal pressed()
+  property bool pressed: false
+  MouseArea
+  {
```
ChALkeR: Please use consistent formatting if possible. E.g that would be `MouseArea {` on lines 7-8.

ChALkeR:
Thanks!

PR #443, opened 2019-02-12
FontLoader.js gives back a font object

wian007:
Tried to use FontLoader with FontLoader@0.0.2.
Found out that the name property wasn't filled and status was 2 ( Loading )
After debugging I found out that the FontLoader@0.0.2 library gives back a font object with the property family instead of a string containing fontFamily.

### DIFF
--- a/src/modules/QtQuick/FontLoader.js
+++ b/src/modules/QtQuick/FontLoader.js
@@ -82,8 +82,8 @@ class QtQuick_FontLoader extends QtQml_QtObject {
             }
           }
         },
-        fontLoaded: fontFamily => {
-          if (this.$lastName === fontName && fontFamily === fontName) {
+        fontLoaded: font => {
+          if (this.$lastName === fontName && font.family === fontName) {
             this.name = fontName;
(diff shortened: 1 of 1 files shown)


src/modules/QtQuick/FontLoader.js
```diff
@@ -82,8 +82,8 @@ class QtQuick_FontLoader extends QtQml_QtObject {
             }
           }
         },
-        fontLoaded: fontFamily => {
-          if (this.$lastName === fontName && fontFamily === fontName) {
+        fontLoaded: font => {
+          if (this.$lastName === fontName && font.family === fontName) {
```
ChALkeR: Perhaps add
```js
const fontFamily = font.family || font; // for various FontLoader.js versions
```
and restore `fontFamily === fontName` in the if block instead?

ChALkeR:
@wian007 Yes, that changed in FontLoader at some point back in 2015: https://github.com/smnh/FontLoader/commit/216b054046d3ee4664585047109394ce304ebe2e.

I guess we could use only the new syntax and ignore older FontLoader versions.

PR #444, opened 2019-02-12
Added vertical align to Text

### LINKED ISSUE #442: qml text in rectangle position bug
Rectangle {
        id: rectangle1
        color: "#ffffff"
        anchors.bottom: rectangle.top
        anchors.bottomMargin: 0
        anchors.top: parent.top
        anchors.topMargin: 0
        anchors.right: parent.right
        anchors.rightMargin: 0
        anchors.left: parent.left
        anchors.leftMargin: 0

        Text {
            id: text1
            text: qsTr("0.1")
            anchors.fill: parent
            verticalAlignment: Text.AlignVCenter
            horizontalAlignment: Text.AlignHCenter
            font.pixelSize: 12
        }
    }

0.1 always on screen top but it is on screen middle in Qt creator

wian007:
I added verticalAlignment to the Text Object. I've used table, table-cell and vertical-align in the css style properties.
That way it seems compatible with the most browsers.

It seems to work with fixed width and height, and with the use of anchors.

I think it also fixes #442

### DIFF
--- a/src/modules/QtQuick/Text.js
+++ b/src/modules/QtQuick/Text.js
@@ -18,6 +18,7 @@ class QtQuick_Text extends QtQuick_Item {
     lineHeight: "real",
     wrapMode: { type: "enum", initialValue: 0 }, // Text.NoWrap
     horizontalAlignment: { type: "enum", initialValue: 1 }, // Text.AlignLeft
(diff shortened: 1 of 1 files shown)


ChALkeR:
Merged and thanks, @wian007! Sorry for the delay on this.

PR #445, opened 2019-04-02
doc: add Wian van Aggelen to AUTHORS

ChALkeR:
/cc @wian007 — is that the preferred name and email?

Sorry for the delay on this.


wian007:
Yes for now it's ok 😊

